PR #20, opened 2017-08-18
[RCL-92] print method for civis_ml

patr1ckm:
This PR adds a print methods showing name, job id, run id, and appropriate metrics for the model.

This is subjective, so feel free to make some suggestions.

screen shot 2017-08-18 at 5 40 08 pm

Details:

- `civis_ml` now has subtype `civis_ml_regressor` and `civis_ml_classifier` (useful for other methods)
- works for multi-class and multi-outcome
- small refactor to group print methods together
- adds some example output (and a generation script) for further development


R/civis_ml_utils.R
```diff
@@ -0,0 +1,45 @@
...
+print.civis_ml_regressor <- function(x, ...) {
+  dv_names <- x$model_info$data$target_columns
+  tab <- with(x$metrics$metrics, t(cbind(mad, rmse, r_squared)))
+  colnames(tab) <- dv_names
+  rownames(tab) <- c("MAD", "RMSE", "R-squared")
+  print.civis_ml(x, tab, ...)
+}
+
+#' @export
+print.civis_ml <- function(x, tab, digits = 4, ...) {
```
wlattner: Does this one need to be exported?

R/civis_ml_utils.R
```diff
@@ -0,0 +1,45 @@
...
+  print.civis_ml(x, tab = tab, ...)
+}
+
+#' @export
+print.civis_ml_regressor <- function(x, ...) {
+  dv_names <- x$model_info$data$target_columns
+  tab <- with(x$metrics$metrics, t(cbind(mad, rmse, r_squared)))
+  colnames(tab) <- dv_names
+  rownames(tab) <- c("MAD", "RMSE", "R-squared")
+  print.civis_ml(x, tab, ...)
```
wlattner: `tab = tab` just for consistency

R/civis_ml_utils.R
```diff
@@ -0,0 +1,45 @@
...
+}
+
+#' @export
+print.civis_ml <- function(x, tab, digits = 4, ...) {
+  wfl <- model_workflow(x)
+  run_id <- x$run$id
+  job_id <- x$job$id
+  cat("<CivisML ", wfl, ">\n", sep = "")
+  cat("Job id: ", job_id, " Run id: ", run_id, "\n", fill = TRUE)
```
wlattner: Maybe we should include the URL of the model too.

patr1ckm:
Added the job URL. This is really nice for a quick copy-paste.

screen shot 2017-08-21 at 9 26 48 am

wlattner:
@keithing and @patr1ckm what do we think about job url vs model url?

keithing:
Hmm, if we return the job id, then the user can find the model url.  If we don't return the results url, is there an easy way for them to find it?

patr1ckm:
N/m, the URL should definitely point to /models/ rather than /scripts

wlattner:
cool, LGTM

patr1ckm:
Now points to the platform model results page. The `regression` summary statistics are now printed in the same order, but have 4 rather than 6 significant digits.

[Here](https://platform.civisanalytics.com/#/models/6903578) is the example link

screen shot 2017-08-21 at 11 40 22 am

wlattner:
LGTM, post updated stat order.

PR #25, opened 2017-08-21
[RCL-101] Enhanced error handling with civis_ml

patr1ckm:
Here are some examples of the error handling with civis_ml (and workflows):

screen shot 2017-08-21 at 3 50 44 pm

screen shot 2017-08-21 at 3 51 31 pm

When the error is printed interactively, the length of the error message is determined by `getOption("warning.length")`, which is 1000 characters:

screen shot 2017-08-21 at 3 50 15 pm

You can also use `get_error` and `fetch_logs` on the error objects themselves. 

Details:
- Low level function `run_model` now throws a `civis_ml_error` when the job has reached a failed state.
- A `civis_ml_error` contains the log within the error message and always prints it.
- `get_error` can be used to get the error data with either `try` or `tryCatch`.
- `fetch_logs` can be used to fetch the log string when the error has been caught with `tryCatch`. You cannot `fetch_logs` on `try`, but you can still get the error messages with `get_error` or just by looking at them.
- S3 methods are exported, otherwise `print.condition` gets called instead of `print.civis_ml_error` for example.


R/civis_ml.R
```diff
@@ -827,6 +829,14 @@ fetch_logs.civis_ml <- function(object, limit = 100, ...) {
   format_scripts_logs(logs)
 }
 
+#' @export
+fetch_logs.civis_ml_error <- function(object, limit = 100, ...) {
+  job_id <- attr(object, "args")$id
+  run_id <- attr(object, "args")$run_id
+  logs <- scripts_list_custom_runs_logs(job_id, run_id, limit = limit)
```
wlattner: Should we add another `tryCatch` here, it's possible to get a 4xx or 5xx error in which case, the job/run wouldn't even exist.

wlattner: I suppose that would hit the `error` branch of the `tryCatch` above on line 537.

patr1ckm: Right. HTTP errors are thrown from `call_api`, caught here in the `tryCatch = error` handler, and re-signalled with `stop`.

wlattner: Got it.

PR #26, opened 2017-08-22
[RCL-94] ENH fetch_predictions

patr1ckm:
I confirmed empirically and via Jamie that the file unload from S3 always has 1 id, no matter the size (checked on a 3M row prediction). The file id points to the manifest, and downloading from the file id automatically downloads all files in the manifest.

Details:
- changed class to `civis_ml_prediction` for consistency, is a sub class of `civis_ml`.
- `fetch_logs` now works for `civis_ml_prediction` 
- added run/job ids and the job url to `print.civis_ml_prediction`

Since we improved the error handling in `civis::run_models`, the error handling in `predict.civis_ml` has improved too, e.g. automatically showing the logs in the error message if the prediction job fails. 

@wlattner


R/civis_ml.R
```diff
@@ -748,17 +750,52 @@ fetch_predict_results <- function(job_id, run_id) {
...
 }
 
+#' Retrieve predictions from a CivisML prediction job as a data frame.
+#' @param x \code{civis_ml_prediction} object from \code{predict.civis_ml}
+#' @param ... arguments passed to \code{read.csv}
+#'
+#' @export
+#' @details
+#' The resulting file can also be downloaded directly using \code{download_civis} (see examples).
+#' @examples
```
wlattner: need `@return`

wlattner: also, will there always be a primary key column?

PR #28, opened 2017-08-23
[RCL-138] Add condition for empty_result_error

keithing:
This is adding behavior like [`civis-python`](https://github.com/civisanalytics/civis-python/blob/master/civis/io/_tables.py#L222) so users can catch the error raised when `read_civis` returns no rows.

### DIFF
--- a/R/io.R
+++ b/R/io.R
@@ -38,11 +38,11 @@
 #'
 #' # Gracefully handle when read_civis.sql returns no rows
 #' query <- sql("SELECT * FROM table WHERE 1 = 2")
-#' tryCatch({
+#' mean_x <- tryCatch({
 #'   df <- read_civis(query, database = "my_database")
-#'   mean_x <- mean(df$x)
+#'   mean(df$x)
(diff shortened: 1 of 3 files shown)


R/io.R
```diff
@@ -35,6 +35,15 @@
...
 #' df <- read_civis(id, using = read.csv)
+#'
+#' # Gracefully handle when read_civis.sql returns no rows
+#' query <- sql("SELECT * FROM table WHERE 1 = 2")
+#' tryCatch({
+#'   df <- read_civis(query, database = "my_database")
+#'   mean_x <- mean(df$x)
+#' }, empty_result_error = function(e) {
+#'   mean_x <- NA
+#' })
```
patr1ckm: What do you think about this example instead? This way `mean_x` gets assigned a value
```
mean_x <- tryCatch({
  df <- read_civis(query, database = "my_database")
  mean(df)
}, empty_result_error = function(e) {
  NA
})
```

PR #35, opened 2017-08-31
[RCL-103] generic function to retrieve logs

patr1ckm:
The `fetch_logs` generic now works for any function that returns a `civis_api` response or `civis_error` from `await`, and an appropriate endpoint exists for fetching the log.

It it currently works for the following functions, which return a `civis_api` object directly:

- `query_civis.numeric` (because it uses `scripts_post_sql`)
- `write_civis`
- `transfer_table`

It would be nice to have this for more functions, but we would need to return the `civis_api` response as an attribute to a given result.

The main win from a user standpoint is that it also works on `civis_error`s if an appropriate log function exists.

So you can program things like: 
```
e <- tryCatch(my_fun, civis_error = function(e) fetch_logs(e))
```
Where the handler writes the log to somewhere, for example.

Details:
- `fetch_logs.civis_api` is exported
- `fetch_logs.civis_error` is exported
- `fetch_logs` docs updated
- `await` now returns more meta data to support `fetch_logs`
- `test_logs.R` tests all `fetch_logs` functionality (including `fetch_logs.civis_ml`)

@wlattner or @keithing for review.


R/logs.R
```diff
@@ -18,6 +26,35 @@ fetch_logs <- function(object, limit = 100, ...) {
...
+
+  # This captures slightly more than necessary, but includes all candidates.
+  available <- grep("*_logs$", ls(getNamespace("civis")), value = TRUE)
+
+  # The only exception this rule is models_list_builds_logs, which we don't use.
+  candidate <- paste0(gsub("get", "list", name), "_logs")
+
+  fetcher <- grep(candidate, available, value = TRUE)
+  if (length(fetcher) == 0) stop("No function to fetch logs.")
+  get(fetcher)
```
keithing: Just a sanity check, but will this work when a function is called with

```r
civis::f(...)
``` 

as well as when a function is called with

```r
library(civis)
f(...)
```

patr1ckm:
```
q <- civis::query_civis("SELECT * FROM datascience.iris")
qrr <- civis::query_civis(q$id)
civis::fetch_logs(qrr)
```
Works!

PR #37, opened 2017-09-01
[RCL-168] civis_ml print method updates

patr1ckm:
Now the print methods work with multitarget multiclass output. I'm punting on multitarget multiclass plots.

screen shot 2017-09-01 at 2 10 50 pm

I also updated the print method for the two class case. Now the AUC is only shown once rather than for each outcome, and the dependent variable name is shown:

screen shot 2017-09-01 at 2 11 05 pm

screen shot 2017-09-01 at 2 11 11 pm

Details
- `hist.civis_ml` and `plot.civis_ml_classifier` now throw an error on multiclass multitarget output.
- `is_multitarget` test for multiclass multitarget output
- the sample models in `generate_civis_ml_test_data.R` have been updated, now use the files endpoint only (run time is ~40s)

@wlattner or @keithing


R/civis_ml_utils.R
```diff
@@ -13,31 +13,65 @@ CIVIS_ML_CLASSIFIERS <- c("sparse_logistic", "gradient_boosting_classifier",
...
+  cat("Job id: ", job_id, " Run id: ", run_id, "\n", fill = TRUE)
+
+  if (any(is_multiclass(x)) & is_multitarget(x)) {
+    tabs <- mapply(class_metric_table, p_cor, class_names, aucs)
+    } else if (is_multiclass(x)) {
+      tabs <- list(class_metric_table(p_cor, class_names, aucs))
+      } else {
+        cat("AUC: ", signif(aucs, digits = digits), fill = TRUE)
+        tabs <- list(class_metric_table(p_cor, class_names))
+      }
```
keithing: small thing, but thought I mention because I was confused for a second:  Can we put the `else if` and `else` on the same indent level as `if` ?

patr1ckm: thanks for catching this

PR #41, opened 2017-09-05
FIX: use package_file to find .rds models in vignette build

patr1ckm:
It compiled in my master branch with `build_site()`, should work in the `gh-pages` branch also. The doc PR may have to be re-run?


patr1ckm:
Looks like we need official 'approval'

PR #42, opened 2017-09-06
[RCL-169] [RCL-171] update to civis 1.0.0 and update changelog

patr1ckm:
This should be good to go for release. `build_win` is clean.

- Client now autogenerates on install in windows magically via `configure.win` 
- The cached spec in `R/sysdata.rda` is now exactly the same as `generated_client.R` that we keep under version control (so no weird overwriting when no API key is present).

Here's a screen shot from the [windows build check](https://win-builder.r-project.org/UmAjK38GC3n2/00install.out) showing the autogeneration:
screen shot 2017-09-10 at 12 30 36 pm

@keithing


patr1ckm:
@keithing, this is ready for review.

keithing:
We should probably wait for #43 to let that change make it into v1.0.0

PR #47, opened 2017-09-15
Updating Docs via Automated Script

keithing:
@patr1ckm I reran adding `utils` and `graphics` to the default packages on `Rscript`.  This should fix the current issues.


patr1ckm:
LGTM! 

In addition to building locally, I also verified that the platform script ran without errors. 

Sorry that this has been so tedious!

keithing:
great, thanks for the thorough review!

PR #48, opened 2017-09-15
[RCL-59] [RCL-60] container backend

patr1ckm:
`test_civis_future.R` shows some usage examples (still needs to be mocked). Basically it's just:
```
plan(civis_platform)
x <- future({2 + 2})
value(x)
```
or
```
plan(civis_platform)
y %<-% {2 + 2}
```

It tries valiantly to detect global variables and install packages too:
```
plan(civis_platform)

library(gbm)
x <- future({
  n <- 500
  x <- rnorm(n)
  y <- x*5
  gbm(y ~ x, data.frame(y = y, x = x), distribution = "gaussian")
})
value(x)

# globals
plan(civis_platform)
a <- 5
x <- future({a + 5})
value(x)
```
@wlattner 
@keithing 

- [ ] print error log on container errors
- [x] mock tests for `r_remote_eval`
- [x] unit tests
- [ ] test `caret` integration


DESCRIPTION
```diff
@@ -18,21 +18,22 @@ LazyData: true
...
     ggplot2,
+    httr,
+    jsonlite,
+    lubridate,
+    methods,
     memoise,
+    purrr,
+    roxygen2,
+    stats,
     stringr,
```
patr1ckm: This just puts the dependencies in alphabetical order

patr1ckm:
@keithing this is ready for review. 

A note that tripped me up for a bit is that `Future` objects are environments. This means that the objects maintain state across function calls. So

- subsequent calls to `value` will just return the stored `value` of the future, not try to fetch the results again.
- A future can only be `run` once. 

I think this is the behavior we want.

I decided to use the platform states `succeeded, failed, cancelled` to indicate the state of the future. The `future` API also adds `running` and `created`. 

I'm leaving off the error handling and `caret` integration for now, this is complex enough.

keithing:
>subsequent calls to value will just return the stored value of the future, not try to fetch the results again.
A future can only be run once.
I think this is the behavior we want.

Agreed.

> I decided to use the platform states succeeded, failed, cancelled to indicate the state of the future. The future API also adds running and created.

Nice.  That should cover all the possible states.

> I'm leaving off the error handling and caret integration for now, this is complex enough.

Yeah, that will make review easier.

R/civis_future.R
```diff
@@ -0,0 +1,167 @@
+#' Evaluate an expression in Civis Platform
+#'
+#' @param ... Arguments to \code{\link{CivisFuture}}
+#' @return The result of evaluating \code{expr}.
+#'
+#' @examples \dontrun{
+#'  eval_civis(quote(sqrt(10)))
```
keithing: should this be `civis_platform` ?

R/civis_future.R
```diff
@@ -0,0 +1,167 @@
...
+#' }
+#'
+#' @export
+civis_platform <- function(...) {
+  future <- CivisFuture(...)
+  if (!future$lazy) future <- run(future)
```
keithing: where does the function `run`  come from?  Does there need to be a `::` or an import?

R/civis_future.R
```diff
@@ -0,0 +1,167 @@
...
+  future
+}
+class(civis_platform) <- c("CivisFuture", "future", "function")
+
+#' Evaluate an expression in Civis Platform
+#' @inheritParams future::Future
+#' @param required_resources resources
```
keithing: Maybe mention these are the resources that get passed into `scripts_post_containers` (or wherever the more complete docs for `required_resources` live).

R/civis_future.R
```diff
@@ -0,0 +1,167 @@
...
+#' @param docker_image_name the image for the container script.
+#' @param docker_image_tag the tag for the Docker image.
+#' @param ... arguments to \code{\link{scripts_post_containers}}
+#'
+#' @return A future evaluating \code{expr} on the given container.
```
keithing: This returns a `future::Future`, right?

R/civis_future.R
keithing: Ah, okay I see this in L72

R/civis_future.R
```diff
@@ -0,0 +1,167 @@
...
+
+fetch_output <- function(run) {
+  outputs <- scripts_list_containers_runs_outputs(run$containerId, run$id)
+  # We only save one object in the run script.
+  output_file_id <- outputs[[1]][["objectId"]]
+  read_civis(output_file_id)
+}
+
+make_docker_cmd <- function(task_file_id, run_script_file_id) {
+  cmd <- "Rscript -e \"civis::download_civis(${run_script_file_id}, 'run.R')\" && Rscript run.R ${task_file_id}"
```
keithing: could this trip anyone up because `Rscript` doesn't import the same libraries as the interpreter?

R/civis_future.R
```diff
@@ -0,0 +1,167 @@
...
+  stringr::str_interp(cmd)
+}
+
+upload_runner_script <- function() {
+  path <- system.file("scripts", "r_remote_eval.R", package = "civis")
+  write_civis_file(path, name = "r_remote_eval.R")
+}
+
+# This is to get around R CMD check
```
keithing: https://github.com/HenrikBengtsson/future/commit/da1d57c1a7fff940b633e5e93b20a959301c9c2b

Maybe we should just rely on the most recent version? Looks like `getGlobalsAndPackages` is now exported.

R/civis_future.R
```diff
@@ -0,0 +1,167 @@
...
+                           docker_image_tag = docker_image_tag, ...)
+  structure(future, class = c("CivisFuture", class(future)))
+}
+
+#' @importFrom future run value resolved
+NULL
+
+#' @export
+run.CivisFuture <- function(future, ...) {
+  cargo <- c(expr = future$expr, envir = future$envir, packages = list(future$packages))
```
keithing: is the package `future` in `list(future$packages)` ?  If so, I wonder if we could strip it out rather than relying on suppressing warnings in the `r_remote_eval.R` script.

inst/scripts/r_remote_eval.R
```diff
@@ -0,0 +1,41 @@
...
+suppressWarnings(fut <- read_civis(task_file_id))
+
+# install and load missing packages
+cat("Installing and loading required packages", fill = TRUE)
+
+pkgs <- fut$packages
+if (length(pkgs) > 0) {
+  pkg_list <- pkgs[!(pkgs %in% rownames(installed.packages()))]
+  if (length(pkg_list) > 0) install.packages(pkg_list, repos = "https://cloud.r-project.org")
+  req_results <- lapply(fut$packages, require, character.only = TRUE, quietly = TRUE)
```
keithing: What is the reason to use `require` and `quietly` here?  Are we safe from package installation errors?

R/civis_future.R
patr1ckm: I'll just move this to the top for clarity

R/civis_future.R
patr1ckm: I'm 100% fixing this, thank you

inst/scripts/r_remote_eval.R
patr1ckm: This just suppresses any package loading messages. Any installation errors would still pop up.

R/civis_future.R
patr1ckm: Now that I think about it, we now have to depend on future for the API. `future` will automatically get loaded with `library(civis)`, so no warning.

DESCRIPTION
```diff
@@ -18,21 +18,22 @@ LazyData: true
...
-    jsonlite,
-    purrr,
-    methods,
-    stats,
-    roxygen2,
     DBI,
     dplyr (>= 0.7.0),
     dbplyr (>= 1.0.0),
+    devtools,
+    future,
```
keithing: Does this need to be `future (>= 1.6.1)`

patr1ckm:
@keithing approval at your leisure?

patr1ckm:
Haha, new acronym? FSTM (feels slow to me)

PR #49, opened 2017-09-18
[RCL-174] fix build issues

patr1ckm:
I think this will fix the [build issues](https://cran.r-project.org/web/checks/check_results_civis.html) on CRAN. 

On unix platforms, we use `bin/sh` instead of `bin/bash` as per [this footnote](https://cran.r-project.org/doc/manuals/r-devel/R-exts.html#FOOT24). `Rscript` is located through the use of the `R_HOME` environment variable temporarily created in `R CMD INSTALL`, as per the examples in the [configure docs](https://cran.r-project.org/doc/manuals/r-devel/R-exts.html#Configure-and-cleanup). 

Windows uses uses the `ash` shell. Like in unix alikes, we use locate `R_HOME` to find `Rscript.exe` (also documented [here](https://cran.r-project.org/doc/manuals/r-devel/R-exts.html#Configure-and-cleanup)).

I'm still working on replicating the environments these were created in. 

@keithing

### DIFF
--- a/configure
+++ b/configure
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
 
-Rscript tools/run_generate_client.R
+"${R_HOME}"/bin/Rscript tools/run_generate_client.R
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,14 @@
 # Changelog
 
+## [1.0.1] - 2017-09-18
+
+### Fixed
+
+- `configure` and `configure.win` now use `R_HOME` to find `Rscript` instead of assuming that `Rscript` is on the path.
+- `configure` uses `bin/sh` instead of `bin/bash`
+- `configure.win` uses `Rscript.exe` for windows builds
+- `configure.win` does not use `bin/bash`.
+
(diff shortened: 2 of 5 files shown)


configure
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
 
-Rscript tools/run_generate_client.R
+${R_HOME}/bin/Rscript tools/run_generate_client.R
```
keithing: Does `${R_HOME}` need to be `"${R_HOME}"`. The docs seem to suggest it does:

https://cran.r-project.org/doc/manuals/r-devel/R-exts.html#Configure-and-cleanup

patr1ckm:
Ah, indeed:

`
(Using ‘${R_HOME}/bin/R’ rather than just ‘R’ is necessary in order to use the correct version of R when running the script as part of R CMD INSTALL, and the quotes since ‘${R_HOME}’ might contain spaces.)
`

patr1ckm:
Let's give it a try

patr1ckm:
Yep, fixed. Thanks.

PR #52, opened 2017-09-25
[RCL-177] FIX test for non-null default db that was not mocked

patr1ckm:


tests/testthat/test_io.R
```diff
@@ -158,9 +158,12 @@ test_that("write_civis.character warns under failure", {
 
 test_that("write_civis fails if no db given and default not provided", {
   options(civis.default_db = NULL)
-  err_msg <- tryCatch(write_civis(iris), error = function(e) e$message)
-  db_err <- tryCatch(get_db(NULL), error = function(e) e$message)
-  expect_equal(err_msg, db_err)
+  with_mock(
```
keithing: 2 questions:

1) Was this working before because the of `option(civis.default_db = NULL)` call on L160?
2) Do we still need L160 now that we're mocking `get_default_database` ?

patr1ckm: 1. they were both 404 errors actually.
2. no

PR #53, opened 2017-10-11
ENH add new parameters to civis_ml training and prediction

elsander:
This PR adds `validation_data` and `n_jobs` as arguments for training, and `cpu`, `memory`, and `disk_space` as arguments for prediction.

### DIFF
--- a/R/civis_ml.R
+++ b/R/civis_ml.R
@@ -68,10 +68,10 @@
 #' @param notifications Optional, model status notifications. See
 #'   \code{\link{scripts_post_custom}} for further documentation about email
 #'   and URL notification.
+#' @param polling_interval Check for job completion every this number of seconds.
 #' @param validation_data Optional, source for validation data. There are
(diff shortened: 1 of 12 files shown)


elsander:
Oops-- I had the arguments in slightly different orders in a couple of methods. That should be fixed now.

Something I noticed, @patr1ckm: `polling_interval` is taken as an input argument to these functions, but isn't passed through to `create_and_run_model`. Is there a reason for that? If not, I can go ahead and add it in this PR, while I'm at it.

patr1ckm:
It's also not passed to `run_model` and then finally to `await`. If you want to fix that here, that would be awesome.

elsander:
Ok, that should do it. I also decided to rename `cpu`, `memory`, and `disk_space` to match the training arguments, for clarity to users.

elsander:
Sure, just pushed the changes!

PR #61, opened 2017-10-30
[RCL-62] only retry on 413, 429, 502, 503, 504

patr1ckm:
`httr::RETRY` now takes a `terminate_on` argument. We can use this to terminate on all codes that don't have to do with connection issues.

I've gone back and forth on the approach, and if it's too clunky, we don't have to merge.


patr1ckm:
CH-CH-CH-CHANGELOG

PR #62, opened 2017-10-31
[RCL-196] implement write_civis.numeric

patr1ckm:
Sorry to slam you here, but this should be implemented too. This is the second half of [#150](https://github.com/civisanalytics/civis-python/pull/150/files) in civis-python.


R/io.R
```diff
@@ -174,13 +177,17 @@ read_civis.sql <- function(x, database = NULL, using = utils::read.csv,
 #' # Create new table, append if already exists
 #' write_civis(df, "schema.my_table", "my_database", if_exists="append")
 #'
-#' # Create new table with defined diskey / sortkeys for speed!
+#' # Create new table with defined diskey / sortkeys for speed
```
keithing: Isn't it a generally accepted rule that "speed" is always followed by "!" ?   ;)

man/publish_addin.Rd
```diff
@@ -0,0 +1,11 @@
...
+% Please edit documentation in R/rstudio.R
+\name{publish_addin}
+\alias{publish_addin}
+\title{Add in to Publish an R Markdown file to Platform Reports from RStudio}
+\usage{
+publish_addin()
+}
+\description{
+Add in to Publish an R Markdown file to Platform Reports from RStudio
+}
```
keithing: I actually think it's a bug that this is generating docs since it's not exposed.  (Although it is "exposed" in the sense that it's a button... so maybe it needs docs anyway).  Why don't we skip adding this now and fix it in another PR?

[here](https://github.com/civisanalytics/civis-r/blob/master/R/rstudio.R#L1) I believe the `#'` should be `#`

patr1ckm: Thanks for the PR!

R/io.R
patr1ckm: we could have two `!!`

PR #63, opened 2017-10-31
[RCL-189] update ids for CivisML 2.0

patr1ckm:
This just updates the civis-ml model ids and re-generates the models for the tests.

Nothing broke! Solid work @elsander!


patr1ckm:
Ah, sure thing. Thanks. Since external release is a month out, I think it's worthwhile.

patr1ckm:
Alright! 
- `CIVIS_ML_TEMPLATE_IDS` is a data frame containing template ids for prediction and training for all versions.
-  `get_predict_template_id` now finds the correct template for and prediction from an existing model
- `get_train_template_id` is used to set the default civisML training template in the package options when loaded. If the most recent version is not available, the previous version is used.

tests/testthat/test_civis_ml.R
```diff
@@ -516,14 +519,15 @@ test_that("passes a manifest file_id", {
...
               primary_key = NULL,
               output_table = NULL,
               output_db_id = NULL,
               if_output_exists = 'fail',
               model_name = "model_task",
               n_jobs = NULL,
-              cpu_requested= NULL,
-              memory_requested= NULL,
+              cpu_requested = NULL,
+              memory_requested = NULL,
```
elsander: Oops, that might have been my fault. Thanks for linting!

tests/testthat/test_civis_ml.R
```diff
@@ -327,21 +328,20 @@ fake_model <- structure(
   class = "civis_ml"
 )
 
+
```
elsander: Stray newline

tests/testthat/test_civis_ml.R
```diff
@@ -362,7 +362,7 @@ test_that("calls scripts_post_custom", {
   )
 
   script_args <- mock_args(fake_scripts_post_custom)[[1]]
-  expect_equal(script_args$from_template_id, 1111)
+  expect_equal(script_args$from_template_id, fake_model$job$fromTemplateId + 1)
```
elsander: This makes me a little nervous because as far as I can tell, it assumes that the prediction template will always be 1 greater than the training template. This is usually the case (we try to do that), but it's not necessarily true, so if you add new templates in a future release, it would break your tests in a surprising way. Can you grab the id from the data frame directly?

patr1ckm: I'm just going to test `get_predict_template_id` once, and just use it everywhere else.

tests/testthat/test_civis_ml.R
patr1ckm: probably Bill

PR #67, opened 2017-11-02
[RCL-200] [RCL-186] update tests, including no validation

patr1ckm:
I think these changes to tests will be easier to understand all at once rather than separately.

Quick summary:

- TST tests now include stacking, perceptron, hyperband, and no validation
- FIX print, plot, hist work for models with no validation
- FIX get_model_data works with no validation (uses `model_info$data` instead of `metrics$data`)
- FIX escaping in hyperband
- FIX spelling error in civis_ml doc


R/civis_ml_utils.R
```diff
@@ -2,15 +2,37 @@
...
-                         "random_forest_classifier",
-                         "extra_trees_classifier")
+CIVIS_ML_CLASSIFIERS <- c("sparse_logistic",
+                          "gradient_boosting_classifier",
+                          "random_forest_classifier",
+                          "extra_trees_classifier",
+                          "multilayer_perceptron_classifier",
+                          "stacking_classifier")
+
+CIVIS_ML_TEMPLATE_IDS <- data.frame(
```
elsander: It looks like this is a duplicate? I also see it defined on line 37.

R/civis_ml_utils.R
```diff
@@ -2,15 +2,37 @@
...
+  id = c(9112, 9113, 9968, 9969),
+  version = c(1.1, 1.1, 2.0, 2.0),
+  name = c("train", "predict", "train", "predict"),
+  stringsAsFactors = FALSE
+)
+
+# returns a version compatible template id for a given training model without API calls.
+get_predict_template_id <- function(m) {
```
elsander: Looks like this is being duplicated too.

tests/testthat/test_civis_ml_utils.R
```diff
@@ -65,25 +64,25 @@ test_that("get_metrics throws error if not model", {
...
 
 test_that("is_multitarget works", {
-  expect_true(is_multitarget(model_list[[11]]))
-  expect_true(is_multitarget(model_list[[12]]))
-  expect_false(is_multitarget(model_list[[2]]))
-  expect_true(is_multitarget(model_list[[12]]) &
-                all(is_multiclass(model_list[[12]])))
+  test <- sapply(model_list, is_multitarget)
+  ans <- sapply(model_list, function(m) length(m$model_info$data$n_unique_targets) > 1)
+  expect_equal(test, ans)
```
elsander: Hooray for refactoring!

patr1ckm:
That definitely answers the "why is this diff weird" question I was having (doh)

PR #57, opened 2017-10-23
[RCL-195] [RCL-178] multipart upload for write_civis_file

patr1ckm:
This is an R translation of the multipart upload in [#150](https://github.com/civisanalytics/civis-python/pull/150) from civis-python. 

While I was making changes to `write_civis_file`, I also changed the default to store files indefinitely. Fixes [#51](https://github.com/civisanalytics/civis-r/issues/51).

I needed to change `read_civis.numeric` to download the file because `httr::content` couldn't guess the file type of multipart uploads correctly. Making this change simplifies the implementation considerably, and makes `read_civis.numeric` work with `readr::read_csv` too.

Details:

- Files larger than 50mb in write_civis_file use a multipart upload
- multipart uploads in parallel by specifying`future::plan`
- All files stored indefinitely by default
- `read_civis.numeric` works with multipart uploads and `readr::read_csv`
- updates the cached spec to include `files_post_multipart` and `files_post_multipart_complete`.

Thanks @jkulzick for leading the way.

PR #70, opened 2017-11-08
[RCL-194] Update default client

keithing:
Rather than relying on `BASE_RESOURCES`, we now have a pre-defined API spec to use as the default spec.  This updates the generation process and regenerates `generated_client.R` based on this change.


keithing:
I ran this through integration tests and it passed those as well.

The workflow after these changes will be call `tools/generate_default_client.R` like

```bash
Rscript tools/generate_default_client.R default_client_spec.json
```

Where we download `default_client_spec.json` from a special internal job that lists the complete set of exposed API endpoints.

keithing:
Two more notes: 

1. I would suggest looking at this one commit by commit, with the first commit containing the new logic and the second just regenerating the client.
2. This will unblock #57

PR #81, opened 2018-01-23
[RCL-209] v1.2.0 with civisML 2.1

patr1ckm:
- Adds the new templates for civisML 2.1 
- smarter civisML template selection
- Bumps version to 1.2.0
- Updates default client
- removed R requirement for hyperband for MLP

Successful check from win-builder [here](https://win-builder.r-project.org/CjaG3dP1a5P5/00check.log)

Here's a notebook testing `get_train_template_id` for the demo user:
https://platform.civisanalytics.com/#/notebooks/3078

The test for `get_train_template_id` now shows how the templates for 1.1 will be used if none of the other templates are available.

Thanks for the quick catch!


R/zzz.R
```diff
@@ -11,15 +11,5 @@ get_train_template_id <- function() {
   latest <- max(versions)
   id <- CIVIS_ML_TEMPLATE_IDS[CIVIS_ML_TEMPLATE_IDS$version == latest &
                               CIVIS_ML_TEMPLATE_IDS$name == "train", "id"]
-
-  # only check update if API key is set
-  if (!inherits(try(api_key(), silent = TRUE), "try-error")) {
-    if (length(scripts_list_custom(id)) == 0) {
-      # use the previous version; assume that only the latest version is in internal release
```
elsander: I don't think we should remove this check. External users won't have access to templates until they get permissions on Wednesday, which means there will be a period where 2.1 is also in "internal release" by default. In fact, we'll have some clients who are on 1.1, some who are beta users on 2.0, and internal users who are on 2.1. So in fact, I think we should make this logic more general, and check against all available versions of CivisML until we find one we can use (e.g, check for 2.1, if fail, check for 2.0, if fail, check for 1.1, etc.). This logic will be useful for any future releases as well, and it's how the python client is handling it now.

R/civis_ml_utils.R
```diff
@@ -20,8 +20,8 @@ CIVIS_ML_CLASSIFIERS <- c("sparse_logistic",
                           "stacking_classifier")
 
 CIVIS_ML_TEMPLATE_IDS <- data.frame(
-  id = c(9112, 9113, 9968, 9969),
-  version = c(1.1, 1.1, 2.0, 2.0),
+  id = c(9112, 9113, 9968, 9969, 10582, 10583),
+  version = c(1.1, 1.1, 2.0, 2.0, 2.1, 2.1),
   name = c("train", "predict", "train", "predict"),
```
elsander: It looks like the `name` column needs and extra `c('train', 'predict')`. This might actually build correctly, because R will loop through the vector again to fill in the column, but it's kind of a hacky behavior on R's part, and I don't think we should rely on it.

patr1ckm:
This is ready for review.

CHANGELOG.md
```diff
@@ -1,6 +1,6 @@
 # Changelog
 
-## [1.2.0] - Unreleased
+## [1.2.0] - 2018-01-23
```
elsander: I would add notes to the changelog on the removed error checking for MLPs, and the new fallback behavior for the template_ids.

R/civis_ml_utils.R
elsander: Oops, I was looking at an old commit. I see now that you updated this.

PR #88, opened 2018-02-28
query_civis_file

### LINKED ISSUE #86: query_civis_file for executing a query and returning the S3 file id
`scripts_get_sql_runs` returns the file id in `$output`, we can utilize this in query civis to automatically return the file ID from an unloaded query. This would likely spare some IO in a few places.

patr1ckm:
Shall we add some new IO? 

- `query_civis_file.sql` runs the sql  and returns the S3 file id
- `query_civis_file.character` exports a `"schema.table"` and returns the S3 file id
- `query_civis_file.numeric` exports results from an existing sql script id and returns the S3 file id

Closes #86


R/io.R
```diff
@@ -609,6 +610,79 @@ query_civis.character <- function(x, database = NULL, verbose = FALSE, ...) {
   await(queries_get, id = q_id, .verbose = verbose)
 }
 
+#' Export results from a query to S3 and return a file id.
+#'
+#' Exports results from a Redshift SQL query, and returns the id of the file on S3 for use
+#' with \code{\link{read_civis}} or \code{\link{download_civis}}.
+#'
+#' @param x "schema.table", \code{sql("query")}, or a query id.
```
keithing: By query id, we mean the id from a scripted_sql job, correct?  This might lead to confusion with the queries endpoint.

R/io.R
```diff
@@ -609,6 +610,79 @@ query_civis.character <- function(x, database = NULL, verbose = FALSE, ...) {
...
+#' @param database string, Name of database where data frame is to be uploaded.
+#' If no database is specified, uses \code{options(civis.default_db)}.
+#' @param job_name string, Name of the job (default: \code{"Civis S3 Export Via R Client"}).
+#' @param hidden bool, Whether the job is hidden.
+#' @param verbose bool, Set to TRUE to print intermediate progress indicators.
+#' @param csv_settings See \code{\link{scripts_post_sql}} for details.
+#' @param ... unused.
```
keithing: why include this if it is unused?

R/io.R
```diff
@@ -609,6 +610,79 @@ query_civis.character <- function(x, database = NULL, verbose = FALSE, ...) {
...
+}
+
+#' @export
+#' @describeIn query_civis_file Export a \code{"schema.table"} to a file id.
+query_civis_file.character <- function(x, database = NULL, job_name = NULL, hidden = TRUE,
+                                       verbose = verbose, csv_settings = NULL, ...) {
+  if (stringr::str_detect(tolower(x), "\\bselect\\b")) {
+    msg <- c("Argument x should be \"schema.tablename\". Did you mean x = sql(\"...\")?")
+    stop(msg)
+  }
```
keithing: This is a good thought.

patr1ckm:
The `Currently ignored` explanation for the `...` argument is required for R CMD CHECK to pass when using S3 dispatch.

R/io.R
patr1ckm: Thanks, we do it for `read_civis` and `download_civis` too.

patr1ckm:
Ready for round 2!

R/io.R
```diff
@@ -609,6 +610,79 @@ query_civis.character <- function(x, database = NULL, verbose = FALSE, ...) {
...
+  cred_id <- default_credential()
+  if (is.null(job_name)) job_name <- "Civis S3 Export Via R Client"
+  run <- start_scripted_sql_job(database = db,
+                                sql = x,
+                                job_name = job_name,
+                                hidden = hidden,
+                                csv_settings = csv_settings)
+  res <- await(scripts_get_sql_runs,
+               id = run$script_id, run_id = run$run_id, .verbose = verbose)
+  res$output[[1]]$fileId
```
keithing: What do you think about returning `res$output[[1]]` instead of `res$output$fileID` here?  I don't have a strong preference, but I wonder if anyone would use the URL instead of the file id?

keithing:
Code looks good, just one final question.

R/io.R
patr1ckm: That's a good question. I made this choice to mirror the output of `write_civis_file`, which I think is handy. If a user wants to do more advanced things, they can always use the endpoints directly.

PR #91, opened 2018-03-06
ENH update README

### LINKED ISSUE #90: Adding an API key to .Renviron is confusing
In the documentation, add instructions about creating the `.Renviron` file, and restarting `Rstudio` if necessary.

patr1ckm:
I hear questions about `.Renviron` fairly frequently and think it's worth being explicit about. 

I clarified updating too. Using `install.packages` only installs (and triggers the updating) if the package has changed on CRAN. Users need to install from github directly in this case, which always triggers the update.

For contribution, I documented the special environment variable `R_CLIENT_DEV` as well as the contribution process.

Details:
- clarifies .Renviron creation when adding a key
- clarifies how to update the package with devtools
- reorders usage, focusing on the high level functions first
- adds more detail on the contribution process

@edakaceli does this address the `.Renviron` confusion? 

Fixes #90 .


edakaceli:
@patr1ckm Yes, the first bullet point covers the .Renviron issue I was experiencing. Thank you!

PR #92, opened 2018-03-07
ENH "dvs_to_predict" option for model predictions

stephen-hoover:
CivisML v2.2 will include the ability for users to select a subset of targets from a multi-target model to write to the scoring table. If users only care about a subset of the model's outputs, they can save time and disk space by only keeping those predictions.

### DIFF
--- a/R/civis_ml.R
+++ b/R/civis_ml.R
@@ -74,6 +74,9 @@
 #' @param dvs_to_predict Optional, For scoring, this should be a vector of column
 #'   names of dependent variables to include in the output table. It must be a
 #'   subset of the \code{dependent_variable} vector provided for training.
+#'   The scores for the returned subset will be identical to the scores which
+#'   those outputs would have had if all outputs were written, but ignoring some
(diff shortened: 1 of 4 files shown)


stephen-hoover:
This is the R client equivalent of civisanalytics/civis-python#241 . Probably I've forgotten something; I haven't worked with the R client before.

R/civis_ml.R
```diff
@@ -71,6 +71,10 @@
 #'   data for validation, and \code{skip}, which skips the validation step.
 #' @param verbose Optional, If \code{TRUE}, supply debug outputs in Platform
 #'   logs and make prediction child jobs visible.
+#' @param dvs_to_predict Optional, For scoring, this should be a vector of column
+#'   names of dependent variables to include in the output table. It must be a
+#'   subset of the \code{dependent_variable} vector provided for training.
```
elsander: Just like the Python client docs, I think it's good to explain that this saves time and disk, and that all scores are calculated either way (so that the model output is not affected, just the user output).

PR #93, opened 2018-03-12
ENH print CivisML warnings when model results fetched

elsander:
In the python client, fetching CivisML models causes CivisML warnings to print to the screen, instead of keeping them buried in the metadata. This PR replicates that behavior in the R client.

Right now, I just have the warnings print without any annotation, even though they are Python warnings printing to an R console. If you think this needs to be clarified for the user, let me know.


patr1ckm:
Do you have an example model to test it on?

elsander:
Here's an example of the output:

image

R/civis_ml.R
```diff
@@ -693,7 +693,11 @@ civis_ml_fetch_existing <- function(model_id, run_id = NULL) {
     metrics <- tryCatch(must_fetch_output_json(outputs, "metrics.json"),
                         error = function(e) NULL)
     model_info <- must_fetch_output_json(outputs, "model_info.json")
+    # re-raise any CivisML warnings
+    tryCatch( {for (warn in model_info$warnings) cat(warn)},
```
patr1ckm: It might be simpler to check whether `length(model_info$warnings) > 0`?

patr1ckm: I'd also suggest using `warning` instead of `cat` so that the warnings can be suppressed/caught by the user.

elsander: I think I would rather keep the python details in the warning message. It makes debugging easier, and I don't want to mislead the user into thinking that the warning is coming from the R client itself.

I'm doing a tryCatch here because I think some older versions of CivisML do not have the warnings in the metadata. I'm not 100% about that, but I think this approach is safer.

I'm using `cat` so that `\n` in the exception will be rendered as a newline, but I can try wrapping it in `warning`.

patr1ckm: SGTM

elsander: Ah, actually, checking the length should work since it's a list. I'll just have to update the mock in one of the tests to return a list.

patr1ckm:
Yes! This is exactly what I wanted as a user. Thank you!

screen shot 2018-03-12 at 5 13 27 pm

PR #95, opened 2018-03-13
ENH add feather support for local dataframes being sent to CivisML

elsander:
This PR adds feather support for newer versions of CivisML. This resolves a bug where users would pass local dataframes with factor-type columns, which in some cases would be written as integers in the csv, causing unexpected ETL behavior in CivisML (i.e., not categorically expanding factors).

The new behavior is as follows:
- for older versions of CivisML, there will be no change.
- for newer versions of CivisML (which support feather), the client will attempt to use feather for local dataframes by default. If the user does not have feather installed, it will fall back to csv, throwing an error if any columns are factors.

This feature matches Python client functionality: https://github.com/civisanalytics/civis-python/pull/200

### DIFF
--- a/R/civis_ml.R
+++ b/R/civis_ml.R
@@ -515,6 +515,10 @@ civis_ml.character <- function(x,
 #' Try to stash a dataframe in feather format, for applicable CivisML versions.
 #' Otherwise, stash as csv. If any columns are factor, require feather format,
 #' to avoid incorrect type inference reading from csv.
+#'
+#' @param x data.frame to stash
+#'
+#' @return file id where dataframe is stored
 stash_local_dataframe <- function(x) {
   # Try to stash a dataframe in feather format.
   tmpl_id <- getOption("civis.ml_train_template_id")
--- a/R/civis_ml.R
+++ b/R/civis_ml.R
@@ -518,7 +518,7 @@ civis_ml.character <- function(x,
 #'
 #' @return file id where dataframe is stored
 stash_local_dataframe <- function(x) {
-  # Try to stash a dataframe in feathenr format.
+  # Try to stash a dataframe in feather format.
   tmpl_id <- getOption("civis.ml_train_template_id")
   tmp_path <- tempfile()
 
--- a/R/civis_ml.R
+++ b/R/civis_ml.R
@@ -520,29 +520,20 @@ civis_ml.character <- function(x,
 #'
 #' @return file id where dataframe is stored
 stash_local_dataframe <- function(x) {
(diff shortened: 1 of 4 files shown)


patr1ckm:
Quick question before I review: What is the unexpected ETL behavior when the factors are written to strings?

elsander:
I corrected my PR comment to clarify.

patr1ckm:
Thanks for clarifying.

What would you think about just `Import`ing feather instead of having a soft dependency? Factor columns are pretty common. As a user, I'd rather just install the dependencies up front then have a dependency failure pop up on a special case.

elsander:
I'm hesitant to add a new hard dependency to the entire R client just for CivisML-- people may use the client for lots of things without touching modeling. The Python client makes all CivisML-specific dependencies soft dependencies to avoid inconveniencing non-CivisML users. My instinct is to match that here, but I can change it if you feel strongly.

patr1ckm:
I understand that decision, but feel differently. This is a breaking change because simple calls like:

```
m <- civis_ml(iris, model_type = "sparse_logistic", dependent_variable = "Species")
```

Will now throw an error without `feather` installed. So upgrading `civis` will cause any code that used `civis_ml` with a local data frame to fail without manually installing `feather`. 

If feather is in `Imports`, upgrading `civis` will automatically install `feather`, and this call will continue to work without any user intervention. I'd prefer if we do that instead, even though it differs from the python client.

keithing:
> I'm hesitant to add a new hard dependency to the entire R client just for CivisML

I agree with this, but I also agree that we can't make any breaking changes.  I think our choices are

1.  Add a warning and allow the undesirable behavior to happen
2. Force installation of feather

Given that `tidyverse` is an accepted practice in R land (which installs *so much code*), I think it's probably okay to add feather as a dependency.  I would lean towards 2 as well.

elsander:
Ok, that argument makes sense to me. Thanks for the discussion, @patr1ckm and @keithing, I'll make the change.

R/civis_ml.R
```diff
@@ -514,6 +512,32 @@ civis_ml.character <- function(x,
...
 
+#' Try to stash a dataframe in feather format, for applicable CivisML versions.
+#' Otherwise, stash as csv. If any columns are factor, require feather format,
+#' to avoid incorrect type inference reading from csv.
+#'
+#' @param x data.frame to stash
+#'
+#' @return file id where dataframe is stored
+stash_local_dataframe <- function(x) {
+  # Try to stash a dataframe in feathenr format.
```
keithing: `feather` instead of `feathenr`?

PR #102, opened 2018-03-28
rm devtools, export fetch_and_generate_client

### LINKED ISSUE #99: remove devtools as a dependency
`devtools` depends on `git2r`, which has required system [dependencies](https://github.com/ropensci/git2r/blob/master/DESCRIPTION#L25). This can cause local installs of `civis-r` to fail, and make installation more difficult.

We can replace `devtools` with just a call to `roxygen` for generating the documentation.

patr1ckm:
This removes devtools as a dependency and updates the default client. Instead, we just use `roxygen2`. This should leave only one system level dependency on `xml2`. In it's place, I use `roxygen2`, `withr`, and `pkgload`, which were previous dependencies of `devtools`. 

The big win is that we no longer depend or import `git2r` which was causing the original issues today.

Additionally, the client generation code is only run if an API key is present. Previously, we did the autogeneration on the default spec, which wouldn't ever add anything to what we ship with the package.

I thought it might be handy to export `fetch_and_generate_client`. 

Finally, I updated the client.

Fixes #99.


R/utils.R
```diff
@@ -134,3 +134,22 @@ toggle_dev <- function() {
     Sys.setenv(R_CLIENT_DEV = "TRUE")
   }
 }
+
+r_env_vars_ <- function () {
```
keithing: How did you decide on these values?  Maybe a comment here would be nice.

patr1ckm:
It looks like this approach isn't going to work because `pkgload` isn't on CRAN. On the off chance that something is weird in my environment, does `roxygen2::roxygenize(".")` work for you? 

I'm getting:

```
Error in .requirePackage(package) : 
  unable to find required package ‘roxygen_devtest’
Calls: <Anonymous> ... makeClassRepresentation -> makeExtends -> .requirePackage
In addition: Warning message:
In library(package, lib.loc = lib.loc, character.only = TRUE, logical.return = TRUE,  :
  there is no package called ‘roxygen_devtest’
Execution halted
```

I traced the error back to `roxygen2::source_package` which adds this weird package to the environment:

`methods::setPackageName("roxygen_devtest", env)`

As far as I can tell, it's unused.

keithing:
> roxygen2::roxygenize(".") work for you?

Same error for me.  There might be some more context here: https://github.com/klutometis/roxygen/issues/595

It looks like the solution was to use `pkgload`, bringing us full circle.

patr1ckm:
Gross. I think we should close this until `pkgLoad` hits CRAN. I think putting all the code in `civis` to simulate loading packages is excessive. 

The approach did work for me locally though, FWIW.

keithing:
Couldn't we just use `devtools` to install `pkgload` ;). I agree that we can wait on this one.

keithing:
@patr1ckm now that `pkgload` do you want to reopen this PR?

PR #103, opened 2018-03-30
FIX: remove linebreaks in parameter descriptions

patr1ckm:
A test is running here:

https://platform.civisanalytics.com/#/scripts/containers/10601923

Since it passes loading the package, we should be good to go.

### DIFF
--- a/R/generate_client.R
+++ b/R/generate_client.R
@@ -317,7 +317,7 @@ is_obj <- function(x) {
 
 get_descr_str <- function(x) {
   desc <- if (!is.null(x$description)) x$description else ""
-  gsub("\n[[:space:]]+", " ", desc)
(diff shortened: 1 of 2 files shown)


R/generate_client.R
```diff
@@ -316,7 +316,8 @@ is_obj <- function(x) {
 }
 
 get_descr_str <- function(x) {
-  if (!is.null(x$description)) x$description else ""
+  desc <- if (!is.null(x$description)) x$description else ""
+  gsub("\n[[:space:]]+", " ", desc)
```
keithing: do we want `*` instead of `+` to match 0 or more instead of 1 or more?

patr1ckm:
Yeah, that's right.

PR #113, opened 2018-05-03
[RCL-211] Remove read-only restriction from read_civis and download_civis

patr1ckm:
Removes the read-only restriction from `read_civis` and `download_civis`. This enables reads from spectrum, which has been added to the smoke tests.

I tested the example in the ticket locally.


patr1ckm:
No, it's used the the DBI connectors.

patr1ckm:
Merged the CHANGELONG conflict, so need re-approval

PR #127, opened 2018-06-06
library(civis) works even if API key has expired

patr1ckm:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Changed
 - `write_civis_file` automatically retries on upload failures.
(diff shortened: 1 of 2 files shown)


patr1ckm:
Using `try(scripts_list_custom(...), ...)` catches errors from expired API keys too. When the error is caught, we fall back to using the latest CivisML template ids (it rea

`api_key()` will successfully return an expired api_key and try to run `scripts_list_custom` (and fail). This causes package loading to fail when you call `library(civis)` with an expired key.

If a user has an expired key (or no key), the next time they call `library(civis)` they'll get the correct templates.

patr1ckm:
No problem! I should have provided context at first. Thanks!

PR #128, opened 2018-06-20
[RCL-215] 1.5.0

patr1ckm:
This updates the default client for the 1.5.0 release.

I added some logic in `tools/generate_default_client.R` to pull down the API spec automatically, and it sets environment variables appropriately. New usage is

```
cd R/tools && Rscript generate_default_client.R
```


patr1ckm:
And [here](https://win-builder.r-project.org/0Z6VPjjiKt3m/) is the successful check from win-builder.

PR #137, opened 2018-08-08
Added coefs function

MadHobbs:
- added `coef.civis_ml` to `civis_ml_utils.R`
- added tests for `coef.civis_ml` to `test_civis_ml_utils.R`
- Updated CHANGELOG
- Added Self as Contributor


NAMESPACE
```diff
@@ -5,6 +5,7 @@ S3method(civis_ml,character)
 S3method(civis_ml,civis_file)
 S3method(civis_ml,civis_table)
 S3method(civis_ml,data.frame)
+S3method(coef,civis_ml)
 S3method(db_query_fields,CivisConnection)
```
MadHobbs: @patr1ckm Does this look right? Is this supposed to be an `S3method` ?

tests/testthat/test_civis_ml_utils.R
```diff
@@ -106,3 +106,24 @@ test_that("get_train_template_id reverts to last id if others not available", {
   expect_equal(id, ans)
 })
 
+test_that("coef.civis_ml returns correct coefficients", {
+  lr <- civis_ml(mtcars,
+                        dependent_variable = "vs",
+                        model_type = "sparse_logistic")
```
keithing: I think this will fail in the test environment because there isn't a `CIVIS_API_KEY`.  Instead of training a new model here, you should be able to use one of the model from `models_list`, which is defined on Line 4 above.

MadHobbs: You're totally right! Thanks Keith. That is what I ended up doing in the latest commit.

MadHobbs:
Oops, I think I know why this failed! Looks like I need to use a UseMethod here with coef!

NAMESPACE
patr1ckm: Oh, `coef` is already a generic method defined in `stats`, we shouldn't re-implement here. See `stats::coef`

R/civis_ml_utils.R
```diff
@@ -193,3 +193,41 @@ is_multiclass <- function(model) {
...
+#' @details Outputs coefficients with names in the style of `stats::coef`
+#' @param model civis_ml_classifier
+#' @return coefs a vector of coefficients or `NULL` if the classifier does not have coefficients
+#' @export
+coef.civis_ml <- function(model) {
+
+  # fields needed aren't present (i.e.: tree-based model)
+  if(is.null(model[["model_info"]][["model"]][["parameters"]][["coef"]])) {
+    return(NULL)
+  }
```
patr1ckm: `} else {`

tests/testthat/test_civis_ml_utils.R
```diff
@@ -106,3 +106,41 @@ test_that("get_train_template_id reverts to last id if others not available", {
   expect_equal(id, ans)
 })
 
+test_that("coef.civis_ml returns correct coefficients", {
```
patr1ckm: For multinomial/multioutput I'd prefer that coef returns a matrix here, one column per dv instead of a vector.  
```
  (Intercept)   (Intercept)   (Intercept)    Unnamed: 0    Unnamed: 0    Unnamed: 0   Sepal.Width   Sepal.Width   Sepal.Width 
   5.02471952   11.42495563 -113.97449115   -0.10176131   -0.06364976    1.12927960   11.54905775   -4.28593365  -34.72930265 
 Petal.Length  Petal.Length  Petal.Length   Petal.Width   Petal.Width   Petal.Width 
 -10.59597673    1.48185815    1.61049283   -5.41703176   -0.29744050   56.52995417 
```

See

```
m <- nnet::multinom(Species ~ Sepal.Length, data = iris)
coef(m)
```
(you'll after to delete the `civis::coef` generic first)

R/civis_ml_utils.R
Zephyraith: I believe I've fixed the issue with re-implementing the generic (see screenshot).  

screen shot 2018-08-22 at 14 26 54

When I do this, the 'export(coef)' line In NAMESPACE goes away but the one you highlighted 'S3method(coef,civis_ml)' does not -- is that ok?

Zephyraith:
The current tests are mostly just copy/paste of the code being tested, which... doesn't seem right... how can I improve that?

keithing:
> The current tests are mostly just copy/paste of the code being tested, which... doesn't seem right... how can I improve that?

Typically what we'd do in a situation like [here](https://github.com/civisanalytics/civis-r/pull/137/files#diff-a67dfbe2934d0dc74d1c57576b397b01R112) is precompute the correct values and cache them, then compare the code to the expected values.  So you might see something like

```r
expected <- c(32, 31, 32)
coefs <- coef(model_list[1])
expect_equal(expected, coefs)
```

The way you actually pre-compute the `expected` values may actually be running the `coef` function itself (and then validating it is correct by looking at them).  The point of the test at that point is just to make sure that if changes are made to the `coef` function later, that it doesn't change the outputs.

Zephyraith:
Cool, I'll do something similar to what you recommended for the feature importance test.

patr1ckm:
> I'm not sure why there are so many changes to the .Rd files (it looks like whitespace change only). Can we find a way to only include the relevant changes to this PR?

I think merging/rebaseing this PR to Master should fix this.

Here's the output for posterity:

screen shot 2018-08-28 at 10 01 39 am

After rebasing this LGTM!

Issue #150, opened 2019-01-08
BUG: write_civis.numeric needs to specify firstRowIsHeader


patr1ckm:
`write_civis.numeric` imports a file id into redshift directly from redshift, however it doesn't expose the argument `firstRowIsHeader` as an advanced option for the 

You can try it by uploading file `26320073`.

For `write_civis.numeric` we actually don't use `...` for anything, and could pass it to imports post syncs.

Issue #149, opened 2018-11-13
ENH: add `escaped` argument to `write_civis`


patr1ckm:
Adding this argument allows for more successful imports easily because `"` can be escaped either by `""` (double quote) or `\` (escaped). 

`write.csv` by default uses `""`, so this feature flag will not lead to an internal inconsistency for csv files written by R destined for Redshift import.

For files generated by other systems, having this feature flag allows for successful imports of `\` escaped files.

A cleaner solution would be to pass `...` in `write_civis` to `imports_post_syncs` rather than `write.csv`, but that's a breaking change.

Issue #130, opened 2018-07-27
ENH: expose the credentials parameter in `write_civis`


patr1ckm:
This gets some use.

PR #151, opened 2019-01-15
Improves write_civis by adding credential_id, headers, and import_arg…

patr1ckm:
Ehance `write_civis` by adding three arguments: `header`, `credential_id`, and `import_args`. This makes no breaking changes, and addresses #150, #149, #130.

While I would prefer not to add arguments, here are the reasons why I added each:

* `header` - Added for equivalence to the python client. Can easily remove though.

* `import_args` - Additional arguments that could possibly be added in the future for `imports_post_files` and `advancedOptions` in `imports_post_syncs`. 

* `credential_id` - While I originally thought that credentials could be passed using `import_args`, it was confusing to read because `write_civis.character` and `write_civis.numeric` use different importing strategies. `write_civis.character` can use `imports_post_files`, which is a convenience for the more generic `imports_post -> imports_post_syncs` that is used in `write_civis.numeric`. Since `credential_id` is an argument to `imports_post`, and an argument to `imports_post_files`, I thought it was less confusing to for `credential_id` to just be a separate argument to `write_civis` rather than passing it in `import_args`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,15 +1,15 @@
 # Changelog
 
-## [1.6.0] - 2018-08-28
-
-### Added
-- `coef.civis_ml` which returns `civis_ml` model coefficients in the format of `stats::coef`
-- added tests for `coef.civis_ml`
-
 ## Unreleased
 
-### Changed
+### Fixed
+- `write_civis.numeric` now correctly syncs with headers [#150].
 
+### Added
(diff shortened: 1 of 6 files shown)


patr1ckm:
Hm, not sure why Travis isn't seeing this branch, need to take a look.

keithing:
Let me know when this is ready to review! I'll take a look as soon as you sign me as reviewer.

PR #155, opened 2019-02-06
Partial: Adding second argument to `await_all`

### LINKED ISSUE #147: ENH: `await_all` should be vectorized over two arguments
Since platform polling calls like `scripts_get_custom_runs` require two arguments (e.g. a job and script id) this function should be vectorized over two arguments not just one. Adding an optional `.y` parameter to `await_all` should do it.

Zephyraith:
Fixes #147 ENH: `await_all` should be vectorized over two arguments

Does not pass existing tests, submitting what I have so far for any feedback.  Replacing lapply() with mapply() seems to have broken something.

### DIFF
--- a/R/await.R
+++ b/R/await.R
@@ -125,30 +125,7 @@ await_all <- function(f, .x, .y = NULL, ...,
   start <- Sys.time()
   fname <- as.character(substitute(f))
 
-  # abstract that last bit so it can handles n number of variables --
-  # in python it would be list(zip(x,y))
-  # zipped_parameters <- if (is.null(.y)) .x else cbind(.x, .y)
-  # zipped_parameters <- if (is.null(.y)) list(.x) else list(.x, .y)
-  # names(zipped_parameters) <- if (is.null(.y)) c(".x") else c(".x", ".y")
   zipped_parameters <- if (is.null(.y)) .x else as.list(paste(.x, .y))
-
-
-
-  # repeat {
-  #   if (is.null(.y)) {
-  #     responses[!called] <- lapply(q[!called], safe_call_once,
-  #                                  f = f, ..., .status_key = .status_key,
-  #                                  .success_states = .success_states,
-  #                                  .error_states = .error_states,
-  #                                  fname = fname)
-  #   } else {
-  #     responses[!called] <- mapply(safe_call_once, .x[!called], .y[!called],
-  #                                  ..., MoreArgs = list(f = f), SIMPLIFY = FALSE,
-  #                                  .status_key = .status_key,
-  #                                  .success_states = .success_states,
-  #                                  .error_states = .error_states,
-  #                                  fname = fname)
-  #   }
 
   repeat {
     responses[!called] <- lapply(zipped_parameters[!called], safe_call_once,
@@ -166,9 +143,8 @@ await_all <- function(f, .x, .y = NULL, ...,
     if (!is.null(.timeout)) {
       running_time <- as.numeric(difftime(Sys.time(), start, units = "secs"))
       if (running_time > .timeout) {
-        args <- c(list(.x), list(.y), list(...))
+        args <- c(list(zipped_parameters), list(...))
         names(args)[1] <- names(formals(f))[1]
-        # names(args)[2] <- names(formals(f))[2]
         status <- unlist(lapply(responses, function(x) get_status(x$response)))
         stop(civis_timeout_error(fname, args, status))
       }
@@ -184,7 +160,7 @@ await_all <- function(f, .x, .y = NULL, ...,
                       ". Retry ", i, " in ", pretty_time, " seconds")
         message(msg)
       }
-      lapply(seq_along(.x), make_msg)
+      lapply(seq_along(zipped_parameters), make_msg)
     }
     Sys.sleep(interval)
     i <- i + 1
--- a/R/await.R
+++ b/R/await.R
@@ -119,13 +119,19 @@ await_all <- function(f, .x, .y = NULL, ...,
                       .timeout = NULL, .interval = NULL,
                       .verbose = FALSE) {
 
+
   responses <- vector(mode = "list", length = length(.x))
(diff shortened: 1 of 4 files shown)


patr1ckm:
`mapply` simplifies the results by default, maybe that's it?

R/await.R
```diff
@@ -124,13 +125,14 @@ await_all <- function(f, .x, ...,
   start <- Sys.time()
   fname <- as.character(substitute(f))
 
-  repeat {
-    responses[!called] <- lapply(.x[!called], safe_call_once,
-                                 f = f, ..., .status_key = .status_key,
-                                 .success_states = .success_states,
-                                 .error_states = .error_states,
-                                 fname = fname)
+  zipped_parameters <- if (is.null(.y)) .x else as.list(paste(.x, .y))
```
keithing: Do we want `as.list(paste(.x, .y))` here or `mapply(list, .x, .y, SIMPLIFY=FALSE)`?  I think the former will concatenate the parameters into a string, while the latter gives us a list of tuples that can be passed to `safe_call_once`.

R/await.R
```diff
@@ -124,13 +125,14 @@ await_all <- function(f, .x, ...,
...
 
+  repeat {
+    responses[!called] <- lapply(zipped_parameters[!called], safe_call_once,
+                                   f = f, ..., .status_key = .status_key,
+                                   .success_states = .success_states,
+                                   .error_states = .error_states,
+                                   fname = fname)
```
keithing: very small nitpick, but the args `f`, `.success_states`, etc should be aligned with `zipped_parameters`.

R/await.R
Zephyraith: I changed it to lapply(strsplit(paste(.x, .y), ' '), as.integer), which might be more what we want...?

> mapply(list, .x, .y, SIMPLIFY=FALSE)
[[1]]
[[1]][[1]]
[1] 1

[[1]][[2]]
[1] 3


[[2]]
[[2]][[1]]
[1] 2

[[2]][[2]]
[1] 4

> lapply(strsplit(paste(.x, .y), ' '), as.integer)
[[1]]
[1] 1 3

[[2]]
[1] 2 4

keithing: Hoho, I think you're right we want the output with the `strsplit`.  We should be able to do that without needing to do the back and forth between strings and integers with `mapply(c, x, y, SIMPLIFY=FALSE)`.  (i.e. using `c` instead of `list` in `mapply`).

tests/testthat/test_await.R
```diff
@@ -156,67 +156,68 @@ test_that("safe_call_once catches civis_await_error", {
...
-  expect_equal(sapply(x, function(x) x$id), 1:2)
+  expect_equal(sapply(x, function(x) x$args), list(c(1, 3), c(2, 4)))
   expect_equal(sapply(x, function(x) x$job_id), rep(1, 2))
 })
 
-test_that("await_all vectorizes over any argument", {
-  x <- await_all(f_rand, .x = 1:2, id = 1)
-  expect_equal(sapply(x, function(x) x$job_id), 1:2)
-})
+# test_that("await_all vectorizes over any argument", {
```
patr1ckm: Should this test be commented out?

R/await.R
```diff
@@ -109,29 +109,37 @@ await <- function(f, ...,
...
   i <- 1
   start <- Sys.time()
   fname <- as.character(substitute(f))
 
+  if (!is.null(.y) & (length(.x) != length(.y))) {
+    error <- c("Lengths of input parameters (.x and .y) are not equal!")
+    stop(error)
+  }
+
+  zipped_parameters <- if (is.null(.y)) .x else lapply(strsplit(paste(.x, .y), ' '), as.integer)
```
patr1ckm: I think we want `mapply(c, .x, .y, SIMPLIFY = FALSE)` here instead of `lapply(strsplit(paste(.x, .y), ' '), as.integer)`

tests/testthat/test_await.R
Zephyraith: Almost certainly not, but I don't understand how it should be modified given we're adding .y .  Any thoughts?

R/await.R
```diff
@@ -119,13 +119,19 @@ await_all <- function(f, .x, .y = NULL, ...,
...
   i <- 1
   start <- Sys.time()
   fname <- as.character(substitute(f))
 
-  zipped_parameters <- if (is.null(.y)) .x else as.list(paste(.x, .y))
+  if (!is.null(.y) & (length(.x) != length(.y))) {
+    error <- c("Lengths of input parameters (.x and .y) are not equal!")
+    stop(error)
+  }
+
```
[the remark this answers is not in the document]
keithing: I actually think this is fine as is.  It's essentially a type error if the lengths don't match, not an error specific to `civis`.

tests/testthat/test_await.R
```diff
@@ -178,22 +178,22 @@ test_that("await_all returns list of completed responses", {
...
-test_that("await_all vectorizes over any argument", {
-  x <- await_all(f_rand, .x = 1:2, .y = 3:4, id = 1)
-  expect_equal(sapply(x, function(x) x$job_id), 1:2)
-  expect_equal(sapply(x, function(x) x$run_id), 1:2)
-})
+# test_that("await_all vectorizes over any argument", {
+#   x <- await_all(f_rand, .x = 1:2, .y = 3:4, id = 1)
+#   expect_equal(sapply(x, function(x) x$job_id), 1:2)
+#   expect_equal(sapply(x, function(x) x$run_id), 1:2)
+# })
```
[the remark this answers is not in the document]
keithing: Rather than change the tests, would you mind adding new tests with `.y` instead?  You could use the tests you've written here, but include them at the bottom of the file in addition to the tests with just `.x` here.  You don't need to test `civis_timeout_error`, etc, just that the new `.y` behavior works as intended.  Unless I'm missing something, you should be able to pass these tests without removing any lines of code.

Issue #100, opened 2018-03-28
ENH: generalize and export fetch output 


patr1ckm:
We have an un-exported utility function `fetch_output`. It seems useful to generalize and expose it to users:

```
fetch_output(job_id, run_id, name = NULL, using = readRDS)
```
Providing the `name` returns an object of that name. The file id is passed to `read_civis` with an appropriate argument to `using`.

Does it matter if `scripts`, `containers` or `custom` is used?

keithing:
Yeah, that would be great.  It would also be really nice if we had a default of `run_id=NULL` and it just grabs the output from the most recent run.

keithing:
This could be pretty simple if we only returned a file id, allowing users to compose function (so they could use `read_civis` or `download_civis` if needed).  The function could simply return all the outputs of script identified by name (defaulting to the most recent run).

```r
>>> job_id <- -999999
>>> x = fetch_output_file_ids(job_id, run_id=NULL)
>>> x

$some_image.png
[1] 39023

$some_data.json
[1] 232

>>> df = read_civis(x, using = jsonlite::fromJSON)
```

patr1ckm:
Ohhhhh I love that!

hinnefe2:
Just chiming in with a +1 for exposing the function - I literally just wrote a function to do this myself.

PR #160, opened 2019-02-27
Civis script interface (#159)

### LINKED ISSUE #159: ENH: read_civis.civis_script
# Read civis interface to scripts

I'm proposing that `read_civis.civis_script` have two additional arguments: `regex` and `using`. This design mimics the current behavior of `read_civis` which is to return values whenever it is unambiguous to do so.

### Default

The default is `using = NULL`, which returns a named list of file ids. This doesn't return values. It does return a helpful result if you don't know what the object types are that are posted in the script.

```
out <- read_civis(civis_script(234)) 
val <- read_civis(out$name, using = reader))
```

```
> out
$name
[1] 1
```

With a `regex`, `read_civis.civis_script` returns a list of ids maching the regex.
```
out <- read_civis(civis_script(234), regex = ‘name’)
# always returns a list
val <- read_civis(out[[1]], using = reader)
vals <- lapply(out, read_civis, using = reader)
```

### using = reader
With `using = reader`, `read_civis` returns the values directly. 

Here `reader` is used to read all outputs:
```
val <- read_civis(civis_script(234), using = reader)
```

A more relevant case is with `regex`, which you can use to filter outputs that match a file extension, and then parse them with the right `using`:

```
val <- read_civis(civis_script(234), regex = ‘.csv’, using = read.csv))
val <- read_civis(civis_script(234), regex = ‘.rds’)
val <- read_civis(civis_script(234), regex = ‘.json’, using = jsonlite::fromJSON)
```

### JsonValues

A really helpful feature of the API is that json values are simply returned as-is in `out$value`. So any non-null `using` can simply return `JsonValues` directly without attempting to parse/download the output.

Thoughts on this @keithing ? Is it too magical?

patr1ckm:
This implements #159. The large number of lines is mostly docs, the implementation wasn't too bad.

There are four new exported functions:
- `civis_script` annotation of an integer as a `civis_script`
- `read_civis.civis_script` reads ids or values from run outputs of a  `civis_script`
- `fetch_output_file_ids` - returns ids of `civis_script`
- `fetch_output` - returns results of `scripts_list_*_runs_outputs`

Other changes:
- `value.CivisFuture` now uses `read_civis.civis_script`, which also adds this code into the smoke tests.
- The `read_civis` docs had references to `READ ONLY` queries which I removed.

Closes #159, #100


patr1ckm:
Alright, this build is passing!

patr1ckm:
One shortcoming is that this assumes that the run outputs are either `Files` or `JSONValues`. Other possible run output types e.g. `Table`, `Report`, `Project` and `Credential` won't work.

R/io.R
```diff
@@ -103,6 +111,31 @@ read_civis.sql <- function(x, database = NULL, using = utils::read.csv,
...
 }
 
+#' @describeIn read_civis Return run output values or file ids of a \code{civis_script} as a list.
+#' @param regex Regex of matching run output names.
+#' @details
+#' If \code{using = NULL}, \code{read_civis.civis_script} will return run output
+#' file ids, including ids of JsonValues.
+#' Otherwise all run outputs matching \code{regex} will be read into memory
+#' with \code{using}. JsonValues are always returned as values if \code{using} is not \code{NULL}.
+#' Results are always a list. If the script has no outputs, the results are an empty list of length 0.
```
keithing: What do you think about an implementation where instead of returning file_ids, this requires `using`?  Then `using=NULL` would correspond to JsonValues.  If a user wants file_ids, then they could use `fetch_output_file_ids`.

keithing: I'm open to suggestions or leaving it as is, but maybe worth another second to think about.

R/scripts.R
```diff
@@ -0,0 +1,86 @@
...
+
+#' Get a script function matching a job type.
+#' @param job output of \code{jobs_get}
+#' @param fun_type one of \code{"logs"} or \code{"outputs"}
+#' @return The correct output or log fetching function
+#' (e.g. \code{\link{scripts_list_containers_runs_logs}}) based on the job type.
+#' @details container and custom scripts both have the same job type, but can be distinguished
+#' by a non-null \code{fromTemplateId}.
+get_script_fun <- function(job, fun_type = c("outputs", "logs")) {
+  fun_type = match.arg(fun_type)
```
keithing: `=` instead of `<-`

R/scripts.R
```diff
@@ -0,0 +1,86 @@
...
+    name <- SCRIPT_MAPPING$name[which(SCRIPT_MAPPING$job_type == job_type)]
+  }
+  fname <- paste0("scripts_list_", name, "_runs_", fun_type)
+  get(fname)
+}
+
+# containers and custom scripts both have type 'ContainerDocker'
+# sql is of type SqlRunner
+# javascript is of type ScriptedSql
+SCRIPT_MAPPING = data.frame(
```
keithing: `=` instead of `<-`.  Have you been writing Python lately ;)

R/io.R
patr1ckm: Yeah! I think that's a clearer way to handle both cases. Especially since we're exporting `fetch_output_file_ids`.

R/scripts.R
patr1ckm: :disappear:

patr1ckm:
Did you have additional thoughts on the case when the job has no run output? I was questioning whether it was more helpful as an error or length 0 list. 

On a second thought, it seems like these functions should throw an error. The user is requesting an output from a job and expecting output to be used later. It could be more helpful as an explicit error rather than a downstream error.

A downstream error would be one of:
```
> list()$a
NULL
> list()[[1]]
Error in list()[[1]] : subscript out of bounds
```
Which could happen anywhere in the script.

keithing:
> On a second thought, it seems like these functions should throw an error. The user is requesting an output from a job and expecting output to be used later. It could be more helpful as an explicit error rather than a downstream error.

Yeah, I agree with this.  It's a little opinionated, but I think it will work best for our users!

patr1ckm:
This makes the suggested changes. `read_civis(civis_script(...))` now returns a named list of JSONValues (if any). `read_civis(civis_script(...), using = fun)` how returns a named list of results from applying `read_civis` to any `File` outputs. If the job has no run outputs, it assertively throws an error message.

R/io.R
```diff
@@ -103,6 +110,32 @@ read_civis.sql <- function(x, database = NULL, using = utils::read.csv,
...
+#' @details
+#' If \code{using = NULL}, \code{read_civis.civis_script}
+#' will return all JSONValues with name matching \code{regex}.
+#' Otherwise all File run outputs matching \code{regex} will be read into memory
+#' with \code{using}.
+#' Results are always a named list. If the script has no outputs, an error will be signalled.
+#' @export
+read_civis.civis_script <- function(x, regex = NULL, using = NULL, ...) {
+  output <- fetch_output(x, regex = regex)
+  if (is.null(using)) {
```
keithing: What happens if a user calls this on a script with a file_id output but doesn't specify `using`?

patr1ckm:
Alright - 

- no errors on empty results
- `using` is required and `using = NULL` returns JSONValues.

PR #162, opened 2019-02-28
write_job_output

patr1ckm:
This adds a function `write_job_output`. Looking up the right function to use for posting run outputs makes it work for both r scripts and containers.


R/scripts.R
```diff
@@ -56,22 +56,49 @@ fetch_output <- function(x, regex = NULL) {
...
+    name <- basename(filename)
+    file_id <- civis::write_civis_file(filename, name = name, expires_at = NULL)
+    job <- jobs_get(job_id)
+    post_output <- get_script_fun(job, verb = "post", fun_type = 'outputs')
+    post_output(id = job_id,
+                run_id = run_id,
+                object_type = 'File',
+                object_id = file_id)
+  } else {
+    return(filename)
```
keithing: I might expect a `NULL` here rather than filename, am I missing something?

PR #163, opened 2019-02-28
Find

patr1ckm:
I didn't originally plan on this, but I've found myself needing it more than once, e.g. interactively hunting for credentials, template ids, etc.

Meet `find`, motivated by the `civis-python` equivalent. It's cool because it iterates through paginated list endpoints, but slightly less cool because `...` needs to be provided to `fn`, so the keys need to be in a list.

We don't want to use the search endpoints because it's probabilistic and not defined for all object types. This will be deterministic for all `list` endpoints.


R/pagination_helpers.R
```diff
@@ -88,3 +88,40 @@ fetch_until <- function(fn, .until, ...) {
...
+#' @return Items from \code{fn} that match all key value pairs in \code{keys}.
+#' @param ... Arguments passed to \code{fn}.
+#' @examples
+#' \dontrun{
+#' find(credentials_list, list(username = 'username', name = 'cred_name')))
+#' find(templates_list_scripts, list(name = 'template name'))
+#'
+#' }
+#' @export
+find <- function(fn, keys, ...) {
```
keithing: if I understand the code correctly, should this be "find_all" instead of "find"?  I.e. you'll loop over all pages regardless if you find a match?

patr1ckm:
And yes, this basically adds a `keys` arguments to `fetch_all`. It's a little helpful for small-ish endpoints like `users_list` and `credentials_list` which aren't provided by search. 

I hadn't used the search endpoints before either, so I tried it for templates. It's nice and fast. Each result gets a `score`, which are used to rank results. Somewhat unhelpfully, e.g. `Untitled Script (Clone of Message Experiment)` gets a score of `2800` but `Message Experiment` only gets a score of `2053`. I'm not sure how the scores change, so the rank of the results is non-deterministic. The results are probably deterministic. I couldn't figure out a query to return an exact match, but that doesn't mean it's not possible.
 
```
search_list_types() # gives the available search types
search_list(query='Message Experiment', type='template_script') # is an example
```

patr1ckm:
I think this isn't worth it to include.

PR #169, opened 2019-04-02
Update CivisFuture to use FutureResult

### LINKED ISSUE #168: DEPRECATION: Use result() method for CivisFuture instead of value()
Hi, author of future here.  Since future 1.8.0 (2018-04-08) we have support for returning FutureResult objects from futures.  A FutureResult object can hold:

* 'value' - the value (of a non-errored evaluation)
* 'stdout' - any captured stdout output
* 'conditions' - any captured conditions, including messages, warnings, and an error (the latter if evaluation resulted in an error)

The default is still that futures return the old-style "value", i.e. the value of the evaluation unless there was a runtime error, then an error object is returned.

All the future backends that I maintain have been migrated to make use of FutureResult since a while.  I'm now planning to deprecate the old-style "value", which CivisFuture is still using.

The outline to make use of FutureResult is:

* Create your Future objects using `future::Future(..., version = "1.8")` (default is `version = "1.7"`).  This will automatically result in a FutureResult object being returned.

* Drop `value.CivisFuture()` - the default `future:::value.Future()` will take care of `value()` calls.

* Instead, implement a `result.CivisFuture()` that receives the FutureResult object.  You can most likely tweak your existing `value.CivisFuture()` to become this new method.  Have a look at [`future.batchtools:::result.BatchtoolsFuture()`](https://github.com/HenrikBengtsson/future.batchtools/blob/0.7.2/R/BatchtoolsFuture-class.R#L280-L311), and [`future.callr:::result.CallrFuture()`](https://github.com/HenrikBengtsson/future.callr/blob/0.4.0/R/CallrFuture-class.R#L123-L149) for examples.  Hopefully, it's not too complicated.

In addition to the deprecation itself, there are several advantages of moving to FutureResult:s, e.g. standard output, messages and warnings produced when evaluating the future expression will be automatically relayed in the main R session in a uniform and consistent manner across future backends (without the future backend maintainer having to do anything).  In the upcoming releases, there will also be support for basic benchmarking of futures and more.

Let me know if you have any questions.

keithing:
Resolves #168.  This does a very light refactor of `CivisFuture`, implementing `result.CivisFuture` instead of `value.CivisFuture`.  Existing calls to `value(CivisFuture(...))` will continue to work, as they will use the default `value.Future` method, which internally will call `result.CivisFuture`.  This puts us in line with the existing API for extending Future backends.


patr1ckm:
I think we'll need to regenerate docs and the NAMESPACE for this too.

patr1ckm:
I'm rebuilding after clearing cache. I heard murmurings on twitter of breaking changes to the travis build process. :/

keithing:
Not sure this is a Travis issue.  I ran `devtools::install(dep=FALSE)` to install, which works, but I'm running into issues with `R CMD build .`.  Investigating.

NAMESPACE
```diff
@@ -466,6 +465,7 @@ export(reports_put_services_shares_groups)
 export(reports_put_services_shares_users)
 export(reports_put_shares_groups)
 export(reports_put_shares_users)
+export(result.CivisFuture)
```
patr1ckm: This isn't right. It should be an `S3method`

Issue #171, opened 2019-04-22
BUG: await_all fails


patr1ckm:
`await_all` is failing this example:

```
ids <- c(23820523, 23820523)
runs <- c(130390319, 130390319)
await_all(scripts_get_r_runs, .x = ids, .y = runs)
```

I think this has to do with passing args to `call_once` and `safe_call_once`. This will require a little refactoring of the existing `call_once` and maybe the tests as well.

@Zephyraith would you be up to taking a look?

Zephyraith:
Can do!

Zephyraith:
![](https://github.trello.services/images/mini-trello-icon.png) [debug civis::await_all()](https://trello.com/c/kTijPuJU/51-debug-civisawaitall)

PR #172, opened 2019-04-26
Debugging `await_all` (#171)

Zephyraith:
Addresses #171,

Implementing named parameters gets the example in the Issue to work, at the expense of the flexibility of the (...) parameter.  
Other functions in the package referencing `await` or `await_all` that relied on the (...) parameter are now breaking -- I have left them like that for now, pending feedback.


R/await.R
```diff
@@ -73,7 +73,7 @@
 #' \item{20-29: 10s - 1m}
 #' }
 #' @seealso \code{\link{get_status}, \link{get_error}, \link{fetch_logs}}
-await <- function(f, ...,
+await <- function(f, .x,
```
patr1ckm: No, we can't do this. This will break user code just as much as it broke ours.

R/await.R
```diff
@@ -112,7 +112,7 @@ await <- function(f, ...,
 #' @param .y a vector of values to be passed to \code{f} (default \code{NULL})
 #' @export
 #' @describeIn await Call a function repeatedly for all values of a vector until all have reached a completed status
-await_all <- function(f, .x, .y = NULL, ...,
+await_all <- function(f, .x, .y = NULL,
```
patr1ckm: `.x` and `.y` should be the first and second arguments to `f`, that is, we pass the arguments by position (see also purrr::map2). The names have to be generic because the first two arguments to a poller are not always `id` `run_id`. See `workflows_get_executions` for an example. This is why `await` uses `...` to begin with.

R/await.R
```diff
@@ -134,8 +134,9 @@ await_all <- function(f, .x, .y = NULL, ...,
   zipped_parameters <- if (is.null(.y)) .x else mapply(c, .x, .y, SIMPLIFY=FALSE)
 
   repeat {
-    responses[!called] <- lapply(zipped_parameters[!called], safe_call_once,
-                                 f = f, ..., .status_key = .status_key,
+    responses[!called] <- lapply(X = zipped_parameters[!called], FUN = safe_call_once,
```
patr1ckm: We may need a slightly different version of `safe_call_once` to get this to work.

R/await.R
```diff
@@ -178,18 +179,19 @@ safe_call_once <- function(...) {
...
 
-# .id is the first argument to f.
-call_once <- function(f, ..., .id = NULL, .status_key = "state",
+# .id is the first argument to f
+call_once <- function(f, ...,
+                      .status_key = "state",
                       .success_states = c("succeeded"),
                       .error_states = c("failed", "cancelled"), fname) {
-  response <- do.call(f, c(.id, list(...)))
+  response <- do.call(what = f, args = sapply(X = ..., FUN = list))
```
patr1ckm: Removing `.id` is right. I think we should be able to do something like

```
f <- function(x, y) c(x, y)
do.call(f, list(1, 2))
```

Note that it won't work if the args are named
```
do.call(f, list(a=1, b=2))
Error in (function (x, y)  : unused arguments (a = 1, b = 2)
```

R/await.R
```diff
@@ -178,18 +179,19 @@ safe_call_once <- function(...) {
...
   status <- response[[.status_key]]
   if (is.null(status)) stop("Cannot find status")
 
   called <- any(status %in% .success_states)
 
   if (any(status %in% .error_states)) {
-    args <- c(.id, list(...))
+    args <- sapply(X = ..., FUN = list)
```
patr1ckm: calling `list(...)` will give you a list of the args. This is the right refactor though, git `.id.` out of there : )

tests/testthat/test_await.R
```diff
@@ -4,14 +4,14 @@ test_that("await captures success_state from status key", {
   f <- function(x) {
     if (x == 0) list(state = "success", y = 2) else list(state = "fail", y = NULL)
   }
-  expect_silent(r <- await(f, x = 0, .status_key = "state",
+  expect_silent(r <- await(f, .x = 0, .status_key = "state",
```
patr1ckm: It's a good rule of thumb that if we make breaking changes to our own package, the user will experience them too. Because this software gets used a lot, we have a strict 'no breaking changes' philosophy.

patr1ckm:
Great work so far! I left some notes about where to head next.

patr1ckm:
Thanks again for taking a look!

PR #174, opened 2019-05-06
FIX: serialization compatibility on R 3.6.0

patr1ckm:
Small change, long story.

The integration tests were failing in `plan(civis_platform)` because they were on R latest. 

The break happened because the serialization format is changing in R 3.6.0, so any `.rds` files generated by R 3.6.0 require at least R 3.5.0 to read. This failed in the `future` test where the `.rds` we upload to platform was generated by R 3.6.0 but read by an older R version prior to R 3.5.0 (e.g. R 3.4.3).

We have R 3.5.2 in datascience-r 2.8.0, so I can just bump the image to fix. I'm tagging the docker image for `plan(civis_platform)` to `2` to decrease errors of this kind in the future.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## [1.7.0] - Unreleased
 
+### Fixed
+
+- Docker images for `civis_platform` are now tagged to `civisanalytics/datascience-r:2` for
+compatibility with R 3.6.0.
+
 ### Added
 
 Utitilies for platform  scripts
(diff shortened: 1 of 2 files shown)


keithing:
> I'm tagging the docker image for plan(civis_platform) to 2 to decrease errors of this kind in the future.

Did you mean

I'm tagging the docker image for plan(civis_platform) to 2 to decrease errors of this kind in `future`. 

?  ;)

Issue #176, opened 2019-05-14
ENH: minimize dependencies


patr1ckm:
We've accrued some dependencies over the course of development that aren't critical. Currently, we have 93 total recursive dependencies. That's too many for an infrastructure package, IMO. Civis employees can see a full dependency report [here](https://platform.civisanalytics.com/spa/#/reports/69004).

Minimizing the dependencies of the package will make it easier to install for our users by requiring a much smaller subset of CRAN. It will also enable creating smaller/leaner production R images for applications beyond the images we've used so far.

The current dependency list is:
```
Imports:
    dplyr (>= 0.7.0),
    dbplyr (>= 1.0.0),
    devtools,
    feather,
    future (>= 1.8.0),
    ggplot2,
    httr,
    jsonlite,
    lubridate,
    methods,
    memoise,
    purrr,
    roxygen2,
    stats,
    stringr,
    testthat,
    utils
Suggests:
    knitr,
    rmarkdown,
    mockery,
    R.utils,
    rstudioapi,
    yaml
```

Proposed list:

```
Imports:
    future (>= 1.8.0),
    httr,
    jsonlite,
    methods,
    memoise,
    stats,
    utils
Suggests:
   ggplot2, (civisML plots) 
   feather, (civisML)
   knitr,
   mockery,
   rmarkdown,
   roxygen2,
   R.utils,
   rstudioapi,
   testthat,
   yaml
```
Removals: `devtools, dplyr,DBI, dbplyr, lubridate, purrr, stringr`. I'm not sure about `stringr`, but it would be nice to remove. 

There are some downstream risk in modifying common environments. In particular, removing dependencies from `civis` modifies the package environment of `datascience-r`. However, stable environment management should be handled in that repo, not this one. For example, `datascience-r` should handle the issue of moving `feather` and `future` to `Suggests` rather than `Imports`.

After this change, the only system dependency will be on `xml2`, which `httr` uses to parse HTML. 

The impact on existing users should be low. Whenever they update R to a new minor version, they may have to manually install e.g. `feather`. 

- [x] Move optional features to suggests
- [x] Remove DBI interface
- [x] Remove lubridate, purrr, stringr
- [x] Remove devtools, move roxygen2 to suggests

PR #178, opened 2019-05-16
Dependencies Part 2: lubridate, DBI interface

patr1ckm:
This removes `lubridate` (only used once) and the entire DBI interface. The version is bumped to `2.0.0`. We can remove `dplyr, dbplyr, DBI` from our dependencies! 

The only function we need from `dplyr` is `sql`. I just copied it into our package. Since we re-exported `sql` from `dplyr` (i.e. people use `civis::sql`) this is not a breaking a change.

Part 2 of #176 .


keithing:
Should this target a 2.0 branch?  Are there any other changes we want to go into 2.0?

patr1ckm:
Done. That's a responsible response to my giddy excitement.

This decreases our recursive dependencies from 92 to 61.

PR #180, opened 2019-05-17
Dependencies part 3: devtools, roxygen2

patr1ckm:
This removes devtools and suggests `roxygen2`. Client autogeneration only happens on source installs and if a valid API key is set.  

Summary:
`install.packages('civis')` will provide a minimal working installation for I/O with Civis Platform.
`install.packages('civis', depends = TRUE)` - adds `feather`, `ggplot2` for CivisML + build tools
`install.packages('civis', depends = TRUE, type = 'source')` - autogenerates new endpoints.

Not requiring the entire `devtools` ecosystem and installing `roxygen2` optionally is a huge boon to reducing dependencies and install time. It reduces our recursive dependencies from 62 to 28.

Moving `roxygen2` to `Suggests` is a small breaking change that will not effect most users. If a source installation is attempted with a valid API key and without `roxygen2` being installed, `civis` will fail to install. This is a build-time error and not a run-time error.

The error message is easy to diagnose, and the builder should be able to deduce the correct course of action (install `roxygen2`). I don't anticipate this happening often, as source installs only happen by default on Linux systems.

Part 3 of #176 .


patr1ckm:
@keithing do you mind taking a look?

PR #181, opened 2019-05-22
Dependencies part 4: remove purrr, stringr

### LINKED ISSUE #176: ENH: minimize dependencies

patr1ckm:
This removes `purrr` and `stringr` from dependencies. 

There is one minor breaking change that should have minimal impact: `fetch_until` (in pagination) now only accepts a function rather than a function and a formula. The original functionality was the only non-trivial functionality provided by either dependency that is not easily replicated in base.

I made an attempt at turning formulas into functions, but it's complicated, and would just be tech debt.  I don't think this is essential functionality for the package, certainly not essential enough to have `rlang` as a dependency.

I think the breaking change will be minimal in scope because I have never received a question about using the pagination functions in 2 years.

The final dependency total is 22, mainly from `httr` and `future`. There are several system dependencies from `httr`. `civis` can now be installed easily in a `rocker/r-ver` (just a base R image) and only takes 691mb compared to the current version on master, which takes ~1gb, a 30% size reduction.

Fixes #176 .


R/civis_ml_plot.R
```diff
@@ -15,8 +15,8 @@ hist.civis_ml <- function(x, name = NULL, ...) {
     names <- get_model_data(x, "target_columns")
   }
   if (is.array(score_array)) {
-    # 3d array to list of 2d arrays
-    score_array <- purrr::array_tree(score_array, margin = 1)
+    score_array <- lapply(apply(score_array, 1, list),
```
patr1ckm: This solution is appropriate for turning the 3d arrays provided by civisML into a list of 2d arrays.

R/pagination_helpers.R
```diff
@@ -50,7 +50,7 @@ fetch_all <- function(fn, ...) {
 #' This is useful when searching for a particular value or record.
 #'
 #' @param fn The API function to be called.
-#' @param .until A function or formula which returns a boolean value.
+#' @param .until A function which returns a boolean value.
```
patr1ckm: This is the breaking change, since the function was exported

R/utils.R
```diff
@@ -116,14 +116,9 @@ camel_to_snake <- function(s) {
   first_cap <- '(.)([A-Z][a-z]+)'
   all_cap <- '([a-z0-9])([A-Z])'
 
-  s <- stringr::str_replace_all(s, first_cap, "\\1_\\2")
-  s <- stringr::str_replace_all(s, all_cap, "\\1_\\2")
-  stringr::str_to_lower(s)
-}
-
-snake_to_camel <- function(s) {
```
patr1ckm: This is unused and unexported

R/civis_future.R
```diff
@@ -181,9 +181,9 @@ fetch_logs.CivisFuture <- function(object, ...){
 
 make_docker_cmd <- function(task_file_id, run_script_file_id) {
   # this loads the same initial packages in the same order as default R.
-  cmd <- "Rscript -e \"civis::download_civis(${run_script_file_id}, 'run.R')\" && \
-    Rscript --default-packages=methods,datasets,utils,grDevices,graphics,stats run.R ${task_file_id}"
-  stringr::str_interp(cmd)
+  sprintf("Rscript -e \"civis::download_civis(%i, 'run.R')\" && \
+    Rscript --default-packages=methods,datasets,utils,grDevices,graphics,stats run.R %i",
+          run_script_file_id, task_file_id)
```
keithing: all about that base

patr1ckm:
Agreed!

PR #182, opened 2019-05-23
write_civis_file.data.frame

### LINKED ISSUE #126: ENH: write_civis_file.data.frame
What would you think about having this generic? It would write the data frame as a csv to S3 by default. Extra args get passed to `write.csv`, and `row.names = FALSE` will be the default.

This would also work for `tbl_df`, `tbl` produced from e.g. `read_csv` since the objects inherit from `data.frame`:

```
xx <- read_csv("~/test/data.csv")
str(xx)
Classes ‘tbl_df’, ‘tbl’ and 'data.frame':	20 obs. of  3 variables:
 $ X1      : int  1 2 3 4 5 6 7 8 9 10 ...
 $ genre_id: int  13 9 4 18 10 16 19 3 8 7 ...
 $ genre   : chr  "Musical" "Fantasy" "Children" "War" ...
 - attr(*, "spec")=List of 2
  ..$ cols   :List of 3
  .. ..$ X1      : list()
  .. .. ..- attr(*, "class")= chr  "collector_integer" "collector"
  .. ..$ genre_id: list()
  .. .. ..- attr(*, "class")= chr  "collector_integer" "collector"
  .. ..$ genre   : list()
  .. .. ..- attr(*, "class")= chr  "collector_character" "collector"
  ..$ default: list()
  .. ..- attr(*, "class")= chr  "collector_guess" "collector"
  ..- attr(*, "class")= chr "col_spec"
```

And 

```
str(dplyr::as.tbl(data.frame(a=1)), 1)
Classes ‘tbl_df’, ‘tbl’ and 'data.frame':	1 obs. of  1 variable:
 $ a: num 1
```
Since we know that `data.frame` is a tabular thing, it makes sense to store it on S3 in a platform independent way by default rather than using the R specific serialization. Some space would be traded off.

An alternative is `feather`, which would preserve factors better. However, the other side still has to know that `feather` is being used, and I'm not sure this really solves the problem.

@keithing for comments.

patr1ckm:
This adds `write_civis_file.data.frame`. It uploads a data frame to S3 as a csv. This is what most users expect in this case. It is easy, and it increases portability across systems. The #1 FAQ for civis-r at this point is 'how do I write a CSV to S3?`. This fixes that question easily. 

It is a breaking change. Scripts that previously used `write_civis_file` to upload data frames had previously updated them as `.rds`. The `read_civis` default would have previously been sufficient to read the file back again. After the change, all data frame uploads with `write_civis_file` will need to be read back as `read_civis(..., using = read.csv)`. 

This will not affect 
1) any object that is not a data frame
2) any file uploaded from a path. For example `write_civis_file("data.csv")` will not break.

This is probably the largest intentionally breaking change of `2.0`. 

Fixes #126.


keithing:
Why not change `read_civis.numeric` to default to `using=read.csv` as well?

patr1ckm:
This won't prevent all breakages, but could minimize breakages if the most common upload to S3 is a data frame. I don't have data on it, but I'm guessing that is true.

patr1ckm:
Back to you!

R/io.R
```diff
@@ -356,24 +361,48 @@ write_civis.numeric <- function(x, tablename, database = NULL, if_exists = "fail
...
+                                        expires_at = NULL,
+                                        row.names = FALSE,
+                                        ...) {
+  with_tempfile(function(tmp_file, ...) {
+    write.csv(x, file = tmp_file, row.names = row.names, ...)
+    write_civis_file.character(x = tmp_file,
+                               name = name,
+                               expires_at = expires_at)
+  })
+}
```
keithing: if I understand `with_tempfile` correctly, should this read:

```R
  with_tempfile(function(tmp_file, ...) {
    write.csv(x, file = tmp_file, row.names = row.names, ...)
    write_civis_file.character(x = tmp_file,
                               name = name,
                               expires_at = expires_at)
  }, name=name, expires_at=expires_at, row.names=row.name, ...)
```
?

Maybe this works as is.

vignettes/data_import_and_export.Rmd
```diff
@@ -76,7 +75,7 @@ from S3 as data frames. IDs are obtained from `write_civis_file`.
 ```{r, eval=FALSE}
 data(iris)
 id <- write_civis_file(iris)
-df <- read_civis(id)
+df <- read_civis(id, using = read.csv)
```
keithing: I don't think this lines needs to change now

R/io.R
patr1ckm: It does work as-is, because those args are stored in the enclosing environment of the temporary function. Too tricky?

keithing: It looks pretty clear to me, just wanted to double check.

PR #183, opened 2019-05-30
Client 2.0

patr1ckm:
This updates the default client. There was a bug in the client generation code that got hit when generating `aliases_get_object_type`. This is fixed and tested in the first commit, with the client regeneration in the second.


R/generate_client.R
```diff
@@ -98,17 +98,27 @@ build_function_args <- function(params) {
...
 
+#' @importFrom utils tail
 build_function_name <- function(verb_name, path_name) {
   parts <- strsplit(path_name, "/")[[1]]
-  parts <- parts[parts != ""]
-  parts <- parts[!startsWith(parts, "{")]
+  parts <- parts[parts != '']
   parts <- gsub("-", "_", parts)
-
+  args <- tail(parts, -1)
```
patr1ckm: This is a transliteration from the python client here: https://github.com/civisanalytics/civis-python/blob/c220a5244573e37f5687138059029f8c0bc7b957/civis/resources/_resources.py#L400

keithing: This is related to https://github.com/civisanalytics/civis-python/issues/298 correct? Does this fix that issue?

NAMESPACE
```diff
@@ -50,6 +50,18 @@ S3method(write_civis_file,default)
 export(CIVIS_ML_CLASSIFIERS)
 export(CIVIS_ML_REGRESSORS)
 export(CivisFuture)
+export(aliases_delete)
+export(aliases_delete_shares_groups)
+export(aliases_delete_shares_users)
+export(aliases_get)
+export(aliases_get_object_type)
```
patr1ckm: @keithing fixed here!

R/generate_client.R
patr1ckm: Yep! It might not be an exact transliteration.

PR #190, opened 2019-06-20
2.0.0 doc update

patr1ckm:
Here's the actual doc update.


patr1ckm:
@keithing sorry for the double bump, but I'd love to send out a release e-mail today with this updated

PR #195, opened 2019-08-15
[SERD-1445] return JSON from run template

arthir:
Modified run_template() so that if specified, the JSON values are returned instead of the associated file ids

### DIFF
--- a/R/scripts.R
+++ b/R/scripts.R
@@ -124,7 +124,7 @@ run_template <- function(id, arguments, JSONValue=FALSE, ...) {
 	return()
      }
      if (length(json_output) > 1) {
-        warning("Error in returning JSON outputs of template run -- too many JSON outputs")
+        warning("More than 1 JSON output for template. Returning only the first one")
      }
      return(json_output[[1]]$value)
 
--- a/R/scripts.R
+++ b/R/scripts.R
@@ -93,9 +93,10 @@ run_civis <- function(expr, ...) {
 #' @param ... additional arguments to \code{scripts_post_custom}
 #' @return If JSONValue is FALSE, File ids of any run outputs are returned.
 #'         If JSONValue is TRUE, JSON values of first JSON run output is returned.
-#'           If there are no JSON outputs, nothing is returned
-#'           If there are more than 1 JSON outputs, error message is printed 
-#'             and nothing is returned.
+#'           If there are no JSON outputs, warning message is printed
+#'             and nothing is returned
+#'           If there are more than 1 JSON outputs, warning message is printed 
+#'             and the first JSON output is returned.
 #'                              
 #' @export
 #' @family script_utils
--- a/R/scripts.R
+++ b/R/scripts.R
@@ -91,8 +91,12 @@ run_civis <- function(expr, ...) {
 #' @param JSONValue bool (default FALSE) If true, returns the 
 #'                  JSON values instead of the file_ids
 #' @param ... additional arguments to \code{scripts_post_custom}
-#' @return If JSONValue is TRUE, File ids of any run outputs are returned.
-#'         If JSONValue is FALSE, JSON values of any run outputs are returned.
+#' @return If JSONValue is FALSE, File ids of any run outputs are returned.
+#'         If JSONValue is TRUE, JSON values of first JSON run output is returned.
+#'           If there are no JSON outputs, nothing is returned
+#'           If there are more than 1 JSON outputs, error message is printed 
+#'             and nothing is returned.
+#'                              
 #' @export
 #' @family script_utils
 #' @examples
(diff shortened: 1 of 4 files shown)


R/scripts.R
```diff
@@ -99,12 +102,24 @@ run_civis <- function(expr, ...) {
...
+#' run_template(id, arguments = list(arg1 = 1, arg2 = 2), JSONValue=TRUE...)
 #' }
-run_template <- function(id, arguments, ...) {
+run_template <- function(id, arguments, JSONValue=FALSE, ...) {
   job <- scripts_post_custom(id, arguments = arguments, ...)
   run <- scripts_post_custom_runs(job$id)
   await(scripts_get_custom_runs, id = job$id, run_id = run$id)
-  fetch_output_file_ids(civis_script(job$id, run$id))
+  if (JSONValue) {
+     output <- fetch_output(civis_script(job$id, run$id))
```
arthir: This already seems to have the JSON values so decided to use this instead of calling json_values_get()

R/scripts.R
```diff
@@ -99,12 +102,24 @@ run_civis <- function(expr, ...) {
...
+     names <- lapply(output, function(o) o$name)
+     file_values <- stats::setNames(lapply(output, function(o) o$value), names)
```
arthir: Also this means that the output should mimic the output of the file_ids look like a JSON with
{
"output": ....,
"output2": .....
}

R/scripts.R
```diff
@@ -99,12 +106,31 @@ run_civis <- function(expr, ...) {
...
+     json_output <- output[sapply(output, function(o) o$objectType=="JSONValue")]
+     if (length(json_output) > 1) {
+        cat("Error in returning JSON outputs of template run -- too many JSON outputs")
+	return()
+     }
+     if (length(json_output) == 0) {
+        cat("Error in returning JSON outputs of template run -- no JSON output")
+	return()
+     }
+     return(json_output[[1]]$value)
```
arthir: Returning something if and only if there is exactly 1 JSON output

R/scripts.R
arthir: Changed so only single JSON output is returned

R/scripts.R
keithing: My instinct here (if `JSONValue=`TRUE`) would be the following:

* If there is 1 JSONValue return that value
* If there is more than 1 JSONValue, return the first and warn that only the first is returned.
* If there are no JSONValues returned, warn that no value was returned and return `list()`

R/scripts.R
```diff
@@ -99,12 +106,31 @@ run_civis <- function(expr, ...) {
...
+run_template <- function(id, arguments, JSONValue=FALSE, ...) {
   job <- scripts_post_custom(id, arguments = arguments, ...)
   run <- scripts_post_custom_runs(job$id)
   await(scripts_get_custom_runs, id = job$id, run_id = run$id)
-  fetch_output_file_ids(civis_script(job$id, run$id))
+  if (JSONValue) {
+     output <- fetch_output(civis_script(job$id, run$id))
+     json_output <- output[sapply(output, function(o) o$objectType=="JSONValue")]
+     if (length(json_output) > 1) {
+        cat("Error in returning JSON outputs of template run -- too many JSON outputs")
```
keithing: `warning` seems more appropriate than `cat` here.

R/scripts.R
arthir: That makes sense as well. I wasn't positive what the best behavior was for case 2. 
In the latest commit, I changed it to the behavior described.

R/scripts.R
arthir: 👍 changed to warning.

arthir:
@keithing Fixed this and ready for re-review

keithing:
(cc @patr1ckm, we're extending your `run_template` function)

R/scripts.R
```diff
@@ -88,8 +88,15 @@ run_civis <- function(expr, ...) {
...
 #' @param arguments list of arguments to the script.
+#' @param JSONValue bool (default FALSE) If true, returns the 
+#'                  JSON values instead of the file_ids
 #' @param ... additional arguments to \code{scripts_post_custom}
-#' @return File ids of any run outputs are returned.
+#' @return If JSONValue is FALSE, File ids of any run outputs are returned.
+#'         If JSONValue is TRUE, JSON values of first JSON run output is returned.
+#'           If there are no JSON outputs, nothing is returned
+#'           If there are more than 1 JSON outputs, error message is printed 
+#'             and nothing is returned.
```
keithing: This is stale since your update to the behavior for > 1 JSON output

arthir: and.. i think we're there now.

R/scripts.R
```diff
@@ -99,12 +107,33 @@ run_civis <- function(expr, ...) {
 #'
 #' # Run the template
 #' run_template(id, arguments = list(arg1 = 1, arg2 = 2), ...)
+#'
+#' # Run the template and return JSON value outputs
+#' run_template(id, arguments = list(arg1 = 1, arg2 = 2), JSONValue=TRUE...)
```
keithing: missed a comma here:

`JSONValue=TRUE, ...`

keithing: Also, for consistency this should be `JSONValue = TRUE`

arthir: fixed it.

arthir:
> Code looks good to me. Like in the python client, could you also add a test and update the Changelog here as well?

Added tests and updated changelog

PR #219, opened 2019-12-30
use credential if provided

emmacooperpeterson:
I may be missing some important context here, but this enables passing a credential to `query_civis()`. Definitely open to discussing other options!


R/io.R
```diff
@@ -703,8 +703,9 @@ query_civis.numeric <- function(x, verbose = FALSE, ...) {
 query_civis.character <- function(x, database = NULL, verbose = FALSE, ...) {
   db <- get_db(database)
   db_id <- get_database_id(db)
-  cred_id <- default_credential()
-  q_id <- queries_post(db_id, x, preview_rows = 0, credential = cred_id, ...)[["id"]]
+  credential <- list(...)$credential
+  credential <- if (is.null(credential)) default_credential() else credential
+  q_id <- queries_post(db_id, x, preview_rows = 0, ...)[["id"]]
```
patr1ckm: Great catch. I think:
`queries_post(db_id, x, preview_rows = 0, credential = credential, ...)[["id"]]`

emmacooperpeterson: actually this was what i had at first, but it didn't work :/

i don't totally understand it, but i think it's because if you call `query_civis()` with a `credential` param, the `credential` is in the `...` – and then when `queries_post()` gets called, it's still getting passed through in the `...`

so passing in `credential = credential`  to `queries_post()` here results in the `formal argument "credential" matched by multiple actual arguments` error

am i following this behavior correctly?

emmacooperpeterson:
@patr1ckm thanks for the test/query_civis_file suggestion ! i'll get started on that

R/io.R
patr1ckm: Oh right! Then `credential` would be in `...` and passed explicitly. Instead we could update the args list. We
```
args <- append(list(...), other, named, args)
# update args$credential
do.call(queries_post, args)
```

patr1ckm:
@emmacooperpeterson is this ready? Thanks again!

emmacooperpeterson:
@patr1ckm i haven't had any time to finish this up 😩i think the last remaining piece is adjusting the test for `query_civis_file`

emmacooperpeterson:
@patr1ckm i _think_ this is ready! the tests are passing but i'm not totally sure i updated them correctly

patr1ckm:
LGTM!

Thanks for contributing, and the tests look 🥇

PR #221, opened 2020-01-22
[WIP] [PARO-741] Use aliases instead of template IDs in the R API client

### LINKED ISSUE #202: ENH: Point civis_ml script templates to aliases
Instead of hard coding CivisML template ids, we can point to aliases (to be developed). The aliases will be updated as part of releases, so if the client points to the “most current version” alias, we won’t need a client release to coordinate with a CivisML release.

This is blocked by CivisML template aliases being released.

sahildshah1:
This PR calls the API, gets the aliases and use the latest template IDs

The changes in this PR were modeled after the changes made to the Python API client to use aliases instead of template  IDS. See https://github.com/civisanalytics/civis-python/pull/341

This PR resolves #202


sahildshah1:
The checks failed.  From the logs it seems like the error is that "The environmental variable CIVIS_API_KEY is not set. Add this to your .Renviron or call Sys.setenv(CIVIS_API_KEY = '<api_key>'"

R/civis_ml_utils.R
```diff
@@ -19,19 +19,93 @@ CIVIS_ML_CLASSIFIERS <- c("sparse_logistic",
...
+
+        job_type_version = get_job_type_version(alias_obj$alias)
+
+        return(list(id = id,
+                    version = job_type_version$version,
+                    name = job_type_version$job_type))
+
+    })
+
+  CIVIS_ML_TEMPLATE_IDS <- data.frame(matrix(unlist(ids),
```
jacksonllee: Not entirely sure, but is R cool with naming intermediate vars in ALL_CAPS? It looks like `CIVIS_ML_TEMPLATE_IDS` was named as such before this PR because it's a constant, but I might be missing something.

R/zzz.R
```diff
@@ -7,23 +7,14 @@
...
-    while (length(scripts_list_custom(id)) == 0 &&
-           i < length(versions)) {
-      # use the previous version; assume that only the latest version is in internal release
-      prev <- versions[length(versions) - i]
-      id <- CIVIS_ML_TEMPLATE_IDS[CIVIS_ML_TEMPLATE_IDS$version == prev &
-                                  CIVIS_ML_TEMPLATE_IDS$name == "train", "id"]
-      i <- i + 1
-    }
-  }
+  CIVIS_ML_TEMPLATE_IDS <- get_template_ids_all_versions()
```
jacksonllee: same comment re: var naming of `CIVIS_ML_TEMPLATE_IDS`

R/civis_ml_utils.R
```diff
@@ -19,19 +19,93 @@ CIVIS_ML_CLASSIFIERS <- c("sparse_logistic",
...
-  id = c(9112, 9113, 9968, 9969, 10582, 10583, 11219, 11220, 11221),
-  version = c(1.1, 1.1, 2.0, 2.0, 2.1, 2.1, 2.2, 2.2, 2.2),
-  name = c("train", "predict", "train", "predict", "train", "predict", "train", "predict", "register"),
-  stringsAsFactors = FALSE
-)
+
+get_template_ids_all_versions <- function(){
+  # Get template IDs for all accessible CivisML versions.
+
+  template_alias_objects <- fetch_until(aliases_list,
```
sahildshah1: `template_alias_objects <- fetch_all(aliases_list,object_type='template_script')` threw an error (See https://github.com/civisanalytics/civis-r/issues/223 for details). Per @patr1ckm comment,  I used `fetch_until(aliases_list, limit=1000, object_type='template_script', function(x) length(x) == 0))` here without a code change

R/civis_ml_utils.R
```diff
@@ -19,19 +19,93 @@ CIVIS_ML_CLASSIFIERS <- c("sparse_logistic",
...
+  }
+
+  return(list(job_type = job_type, version = version))
+
+}
 
 # returns a version compatible template id for a given training model without API calls.
 get_predict_template_id <- function(m) {
   train_id <- m$job$fromTemplateId
   this_version <- CIVIS_ML_TEMPLATE_IDS[CIVIS_ML_TEMPLATE_IDS$id == train_id, "version"]
```
jacksonllee: If the module-level constant `CIVIS_ML_TEMPLATE_IDS` is removed above, how is this line still operating without crashing while trying to access `CIVIS_ML_TEMPLATE_IDS`? What do I miss?

R/civis_ml_utils.R
```diff
@@ -19,19 +19,93 @@ CIVIS_ML_CLASSIFIERS <- c("sparse_logistic",
...
+
+  names(CIVIS_ML_TEMPLATE_IDS) <- c("id", "version", "name")
+  CIVIS_ML_TEMPLATE_IDS[,"id"] <- as.integer(CIVIS_ML_TEMPLATE_IDS[,"id"])
+
+
+  return(CIVIS_ML_TEMPLATE_IDS)
+
+}
+
+get_job_type_version <- function(alias){
```
jacksonllee: Should we write tests for this function? For civis-python, this is what I did:

https://github.com/civisanalytics/civis-python/pull/341/files#diff-945a68cf4837987241a1965da63ad812R696-R718

R/civis_ml_utils.R
```diff
@@ -19,19 +19,93 @@ CIVIS_ML_CLASSIFIERS <- c("sparse_logistic",
...
                           "stacking_classifier")
 
-CIVIS_ML_TEMPLATE_IDS <- data.frame(
-  id = c(9112, 9113, 9968, 9969, 10582, 10583, 11219, 11220, 11221),
-  version = c(1.1, 1.1, 2.0, 2.0, 2.1, 2.1, 2.2, 2.2, 2.2),
-  name = c("train", "predict", "train", "predict", "train", "predict", "train", "predict", "register"),
-  stringsAsFactors = FALSE
-)
+
+get_template_ids_all_versions <- function(){
```
jacksonllee: Should we write tests for this new function? In civis-python, this is what I did:

https://github.com/civisanalytics/civis-python/pull/341/files#diff-945a68cf4837987241a1965da63ad812R721-R726

sahildshah1: Should these functions be documented per http://r-pkgs.had.co.nz/man.html#man-functions

R/civis_ml_utils.R
sahildshah1: Should this functions be documented per http://r-pkgs.had.co.nz/man.html#man-functions

R/civis_ml_utils.R
```diff
@@ -19,19 +19,93 @@ CIVIS_ML_CLASSIFIERS <- c("sparse_logistic",
...
 
 # returns a version compatible template id for a given training model without API calls.
 get_predict_template_id <- function(m) {
   train_id <- m$job$fromTemplateId
   this_version <- CIVIS_ML_TEMPLATE_IDS[CIVIS_ML_TEMPLATE_IDS$id == train_id, "version"]
-  CIVIS_ML_TEMPLATE_IDS[CIVIS_ML_TEMPLATE_IDS$version == this_version &
-                        CIVIS_ML_TEMPLATE_IDS$name == "predict", "id"]
+
+  #  `prod` points to the same template id as the latest version (e.g. v2.2)
+  CIVIS_ML_TEMPLATE_IDS[CIVIS_ML_TEMPLATE_IDS$version == this_version[1] &
```
sahildshah1: Is this too kludgey to address the fact that `prod` points to the same template id as the latest version (e.g. v2.2) so that when `train_id` corresponds to the `prod` template id `this_version` will be a vector equal to e.g. `"prod", "v2.2"`)

R/civis_ml_utils.R
patr1ckm: Agreed, since this isn't a package level constant anymore, it doesn't need to be in all caps. Not a syntax requirement, but a nice convention for readability.

R/zzz.R
patr1ckm: yep, in can be lower case now since it's not referring to global data

R/civis_ml_utils.R
patr1ckm: +1

R/civis_ml_utils.R
patr1ckm: Doesn't need to be in all caps

R/civis_ml_utils.R
patr1ckm: I would imagine this is crashing : )

This should be refactored to fetch or re-use pre-fetched civis ml template ids.

R/civis_ml_utils.R
patr1ckm: Since it's not user facing, it doesn't need to be extensive, but a few docs would be great.

patr1ckm: I think we should take this opportunity to refactor global package state (including options) and just call the function directly when we need it.

R/civis_ml_utils.R
patr1ckm: Grabbing `prod` by name is clearer to read I think.

R/civis_ml_utils.R
sahildshah1: Changed to lower case. Thanks! Resolving this convo

R/zzz.R
sahildshah1: Changed to lower case. Thanks! Resolving this convo

R/civis_ml_utils.R
sahildshah1: Thanks for calling this out! I added a line to fetch the civis ml template ids.

R/civis_ml_utils.R
sahildshah1: Add roxygen comments for this function. Thanks!

R/civis_ml_utils.R
sahildshah1: Added roxygen comments for this function.

sahildshah1: Thanks for the call out! I wrote a test modeled after the civis-python test. This is my first experience writing tests in R and I used http://r-pkgs.had.co.nz/tests.html as a reference. I ran the test I wrote from the command line and all the expectations were met.

R/civis_ml_utils.R
sahildshah1: Sorry I'm not sure if I fully understand your comment..

Stepping through this function 

 `train_id <- m$job$fromTemplateId` is grabbed from the argument `m`

`civis_ml_template_ids <- get_template_ids_all_versions()` produces

```
civis_ml_template_ids
      id version         name
1  11219    prod     training
2  11220    prod   prediction
3  11221    prod registration
4  10615     dev     training
5  10616     dev   prediction
6  11028     dev registration
7  11219    v2.2     training
8  11220    v2.2   prediction
9  11221    v2.2 registration
10  7020    v0.5     training
11  7021    v0.5   prediction
12  8387    v1.0     training
13  8388    v1.0   prediction
14  9112    v1.1     training
15  9113    v1.1   prediction
16  9968    v2.0     training
17  9969    v2.0   prediction
18 10582    v2.1     training
19 10583    v2.1   prediction
```

If `train_id = 11219` then 

`this_version <- civis_ml_template_ids[civis_ml_template_ids$id == train_id, "version"]` produces 

`"prod" "v2.2"`

Is your suggestion to just return the `prediction` template id for `prod` and not based off of the the argument `m` . Would we have to refactor this when we make changes to make the version run a parameter?

R/civis_ml_utils.R
sahildshah1: Sorry I'm not sure what you mean by "global package state (including options) we can just call the function directly when we need it" I tried to read up on it but couldn't find a specific part of http://r-pkgs.had.co.nz/ to read? Would you be able to point me to where I can learn more about this to make this change? Thanks!

patr1ckm:
So instead of setting a default training template when the package loads here: https://github.com/civisanalytics/civis-r/blob/master/R/zzz.R#L3

We should remove that (don't try to set that package option) and fetch the templates when either the training or scoring jobs are kicked off. This should let the build pass. 

"global package data" just meant that the ids were previously stored in a global variable. It's not R specific. We should remove that. It will require an additional change in how the template ids are obtained in the `civis_ml` file.

R/civis_ml_utils.R
sahildshah1: Okay. I kept it as indexing by integer! Thanks

sahildshah1:
@patr1ckm I refactored to remove ids stored in a global variable. See the [latest commit  ](https://github.com/civisanalytics/civis-r/pull/221/commits/dbf5ab723a5c605fce09104661d1b6fbec26d029)

Per the [instructions in the README ](https://github.com/civisanalytics/civis-r#contributing), I ran  
-- `devtools::test()` locally and all the unit tests passed
-- `devtools::check()` locally and R CMD check passed
-- `Rscript smoke_test.R` and the smoke test passed 

However, the checks still failed. Upon inspecting the logs they said that R CMD check failed. I dug into the logs more and saw another message that "The environmental variable CIVIS_API_KEY is not set"  Do you have any advice on how to get started fixing this? 

cc. @jacksonllee for viz

tests/testthat/test_civis_ml_utils.R
```diff
@@ -95,23 +114,16 @@ test_that("is_multitarget works", {
...
+  ver <- civis_ml_template_ids[civis_ml_template_ids$id == id, "version"]
+  pred_id <- civis_ml_template_ids[civis_ml_template_ids$version == ver[1] &
+                                   civis_ml_template_ids$name == "prediction", "id"]
   expect_equal(get_predict_template_id(m), pred_id)
 
   fake_model <- list(job = list(fromTemplateId = 9112))
   expect_equal(get_predict_template_id(fake_model), 9113)
 })
 
-test_that("get_train_template_id reverts to last id if others not available", {
```
sahildshah1: I deleted this test b/c per @jacksonllee comment in PARO-741

Question: Can we assume the user will always have access to the latest version of CivisML?

Yes. All CivisML aliases for released versions of CivisML are public to all platform users.

R/civis_ml_utils.R
```diff
@@ -19,19 +19,111 @@ CIVIS_ML_CLASSIFIERS <- c("sparse_logistic",
...
+
+  return(id)
+
+}
+
+
+#' Get template IDs for all accessible CivisML versions.
+#'
+#' @return  A data frame containing the template id, CivisML version, and job type.
+get_template_ids_all_versions <- function(){
```
jacksonllee: The function `get_template_ids_all_versions` is called at various places downstream (e.g., inside `get_train_template_id` and `get_predict_template_id`). Is there a convenient way to cache whatever `get_template_ids_all_versions` returns, something similar to [`lru_cache` used in civis-python](https://github.com/civisanalytics/civis-python/blob/e5787ed84d7a78904ca9d9400b070a6ac2bd275d/civis/ml/_model.py#L299)? It'd be nice to avoid making multiple API calls when calling `get_template_ids_all_versions` multiple times in the same runtime session, since the results should be the same.

patr1ckm:
@jacksonllee is right, we'll need to mock out the API calls in relevant tests. 

What I'd probably do is something like:
```
test_data = data.frame(... looks like the output of get_template_ids_all_versions ...)
id <- with_mock(
  `civis::get_template_ids_all_versions` = test_data,
   get_train_template_id()
)
expect_equal(id, 234)
```

You'll need to mock out `get_train_template_id` in the tests of the `civis_ml` functions as well.

R/civis_ml_utils.R
patr1ckm: Yep.

Adding the following should work:
```
get_template_ids_all_versions <- memoise::memoise(get_template_ids_all_versions)
```

sahildshah1:
@patr1ckm Mocked out API calls per our convo this morning, added a test for `get_template_ids_all_versions` and modified the test for `get_predict_template_id` 

There are no longer any failing tests on Travis CI but there are warnings. If i'm reading the logs correctly the warnings are being treated as errors and causing the checks to fail.

```
> test_check("civis")
══ testthat results  ═══════════════════════════════════════════════════════════
[ OK: 1048 | SKIPPED: 0 | WARNINGS: 15 | FAILED: 0 ]
> 
> proc.time()
   user  system elapsed 
  4.373   0.240   6.291 
Found warnings, treating as errors (as requested).
```

tests/testthat/test_civis_ml_utils.R
```diff
@@ -22,6 +22,66 @@ str_detect_multiple <- function(string, pattern) {
...
+                                           displayName = "Model Training - DEV ONLY"),
+                                      list(id = 26,
+                                           objectId = 11221,
+                                           objectType = "template_script",
+                                           alias = "civis-civisml-registration-v2-2",
+                                           userId = 1750,
+                                           displayName = "Trained Model Registration, v2.2"))
+
+  with_mock(
+     `civis::fetch_until` = function(...) fake_template_alias_objects,
```
sahildshah1: @patr1ckm I mocked out `fetch_until` here b/c `template_alias_objects` in the code is returned by `fetch_until`: 
```
  template_alias_objects <- fetch_until(aliases_list,
                                       limit=1000,
                                       object_type='template_script',
                                       function(x) length(x) == 0)
```
When I tried mocking out   out `aliases_list` with the line per our discussion on line yesterday

`civis::aliases_list = function(...) fake_template_alias_objects` 

produced

```
civis_ml_utils.R:39: error: get_template_ids_all_versions
argument is of length zero
Backtrace:
 1. civis::get_template_ids_all_versions()
 2. civis::fetch_until(...) R/civis_ml_utils.R:39:2
 3. civis::next_page(res) R/pagination_helpers.R:82:4
```

Am I okay to mock out `fetch_until` instead of `aliases_list`?

patr1ckm: Yep!

patr1ckm:
We don't need to silence all warnings, just the one warning generated by R CMD CHECK:

```
Codoc mismatches from documentation object 'stash_local_dataframe':
stash_local_dataframe
  Code: function(x, tmpl_id)
  Docs: function(x)
  Argument names in code not in docs:
    tmpl_id
```

Which just means we need to document `tmpl_id` here.

jacksonllee:
LGTM -- everything looks reasonable to me for CivisML. Deferring to @patr1ckm for R / civis-r specific aspects.

patr1ckm:
After making sure the functions run in practice, I'm good to go

sahildshah1:
@patr1ckm @jacksonllee I ran civis_ml  (and `predict`) interactively

```
library(devtools)
load_all()
```

```
gb_model <- civis_ml_gradient_boosting_classifier(iris, "Species")
```

produced

https://platform.civisanalytics.com/spa/#/models/60749121

and

```
predict(gb_model, newdata = iris)
```

produced 

https://platform.civisanalytics.com/spa/#/scripts/custom/60753188

I didn't split the iris data set into testing and training sets and evaluate the results but the functions ran in practice. Is this sufficient testing of the functions?

PR #235, opened 2020-04-19
replaced single-shot HTTP requests with RETRY logic

jameslamb:
In this PR, I'd like to propose swapping out calls to `httr::POST()` etc. with `httr::RETRY()`.  It looks like you're already using `RETRY()` in a lot of places, which is awesome! This PR just changes the few remaining places where you weren't.

This will make the package more resilient to transient problems like brief network outages or periods where the service(s) it hits are overwhelmed. In my experience, using retry logic almost always improves the user experience with HTTP clients.

I also noticed that you are mocking `httr::VERB()` but that it's never called in the code, so I removed those mocks.

I'm working on https://github.com/chircollab/chircollab20/issues/1 as part of Chicago R Collab, an R 'unconference' in Chicago.

Thanks for your time and consideration!

### DIFF
--- a/R/io.R
+++ b/R/io.R
@@ -258,7 +258,7 @@ write_civis.character <- function(x, tablename, database = NULL, if_exists = "fa
           credential_id = credential_id),
       import_args)
     job_r <- do.call(start_import_job, args)
-    put_r <- httr::RETRY("PUT", job_r[["uploadUri"]], body = httr::upload_file(x), terminate_on = c(403, 404))
+    put_r <- httr::RETRY("PUT", job_r[["uploadUri"]], body = httr::upload_file(x), terminate_on = setdiff(400:499, 429))
     if (put_r$status_code != 200) {
       msg <- httr::content(put_r)
       stop(msg)
@@ -564,7 +564,7 @@ download_civis.numeric <- function(x, file,
   args <- c(list(files_get(x)$fileUrl),
             list(httr::write_disk(file, overwrite = overwrite)),
             verb = "GET",
-            terminate_on = c(403, 404))
+            terminate_on = setdiff(400:499, 429))
   if (progress) args <- c(args, list(httr::progress()))
   resp <- do.call(httr::RETRY, args)
 
@@ -904,7 +904,7 @@ download_script_results <- function(script_id, run_id,
   stop_if_no_output(script_results)
 
   url <- script_results[["output"]][[1]][["path"]]
-  args <- list(url, verb = "GET", terminate_on = c(403, 404))
+  args <- list(url, verb = "GET", terminate_on = setdiff(400:499, 429))
   if (!is.null(filename)) {
     args <- c(args, list(httr::write_disk(filename, overwrite = TRUE)))
   }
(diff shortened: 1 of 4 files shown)


R/io.R
```diff
@@ -258,7 +258,7 @@ write_civis.character <- function(x, tablename, database = NULL, if_exists = "fa
           credential_id = credential_id),
       import_args)
     job_r <- do.call(start_import_job, args)
-    put_r <- httr::PUT(job_r[["uploadUri"]], body = httr::upload_file(x))
+    put_r <- httr::RETRY("PUT", job_r[["uploadUri"]], body = httr::upload_file(x), terminate_on = c(403, 404))
```
mheilman, line 261: I think it'd be better to terminate on any 4xx error other 429 here (e.g., `terminate_on  = setdiff(400:499, 429)`).  That's what AWS recommends [here](https://docs.aws.amazon.com/general/latest/gr/api-retries.html), and this URI and the one for `write_disk` below will be S3 URLs.

jameslamb: sure I can do that! `setdiff()` is a super clean way to express it, I like it 😀

jameslamb: addressed the comments in https://github.com/civisanalytics/civis-r/pull/235/commits/7cdb800edc26da45911089a4bab13e3ee9709a0f

R/io.R
```diff
@@ -562,9 +562,11 @@ download_civis.numeric <- function(x, file,
   }
 
   args <- c(list(files_get(x)$fileUrl),
-            list(httr::write_disk(file, overwrite = overwrite)))
+            list(httr::write_disk(file, overwrite = overwrite)),
+            verb = "GET",
+            terminate_on = c(403, 404))
```
mheilman, line 567: Could you also please make this `terminate_on = setdiff(400:499, 429)`?  Sorry, I should have commented about each case rather than just the first.

R/io.R
```diff
@@ -902,14 +904,14 @@ download_script_results <- function(script_id, run_id,
   stop_if_no_output(script_results)
 
   url <- script_results[["output"]][[1]][["path"]]
-  args <- list(url)
+  args <- list(url, verb = "GET", terminate_on = c(403, 404))
```
mheilman, line 907: Could you also please make this `terminate_on = setdiff(400:499, 429)`?

R/io.R
jameslamb: oh sure, no problem!

jameslamb:
I pushed https://github.com/civisanalytics/civis-r/pull/235 to address the most recent review comments

mheilman:
@jameslamb, would you care to merge this? (no rush, I just noticed this was still open)

jameslamb:
> @jameslamb, would you care to merge this? (no rush, I just noticed this was still open)

@mheilman I don't have permission to merge pull requests in this repository. That has to be done by a maintainer.

image

mheilman:
Oh, my bad, sorry. I forgot about that! I thought you could merge if it was approved by a maintainer or something.

PR #238, opened 2020-06-11
[APIC-297] Fix smoke test; update CivisFuture default docker image tag

patr1ckm:
This fixes the failing smoke test, replacing the test with something simpler. 

The default `CivisFuture` image tag is now also latest, since `datascience-r` will be bumping a major version for an R 4.0.0 release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Changed
 - Failed HTTP requests are now retried before raising errors (#235)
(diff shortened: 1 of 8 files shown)


patr1ckm:
It's actually looking like a bunch of stuff isn't passing locally for me on R 4.0.0, may need to take a deeper look.

patr1ckm:
Alright, @mheilman this is ready.

PR #239, opened 2020-06-17
[APIC-298] 3.0.0

patr1ckm:
Ok! This should be ready for the 3.0.0 release. I ran through the devtools release checks and it passes on win-builder (the CRAN windows machine).

Screen Shot 2020-06-22 at 9 47 39 AM

A small update was required to the api generation tests that depended on endpoints `models_, predictions_` that were removed since the last release. 

In addition, I discovered that the file depended heavily on the cached spec from two years ago (I can thank myself for that) so just pinned the tests to that spec.


tests/testthat/test_autogen.R
```diff
@@ -2,13 +2,9 @@
...
-if (file.exists("data/full_spec.rds")) {
-  spec <- readRDS("data/full_spec.rds")
-} else {
-  spec <- suppressMessages(get_spec())
-  saveRDS(spec, "data/full_spec.rds")
-}
-
+# 2020-06-17: this is just a cached spec from a few years ago, updating will require
+# refactoring the tests.
+spec <- readRDS("data/full_spec.rds")
```
patr1ckm, line 7: Just use the cached spec because of all the changes required.

tests/testthat/test_autogen.R
```diff
@@ -2,13 +2,9 @@
...
 
-if (file.exists("data/full_spec.rds")) {
-  spec <- readRDS("data/full_spec.rds")
-} else {
-  spec <- suppressMessages(get_spec())
-  saveRDS(spec, "data/full_spec.rds")
-}
-
+# 2020-06-17: this is just a cached spec from a few years ago, updating will require
+# refactoring the tests.
```
mheilman, line 6: would you mind filing an issue about this?

patr1ckm:
You mentioned that:

> I think changing the default to "latest" qualifies as a breaking change, so I think we should plan for a major version bump

This is a breaking change in default environment though there are no breaking changes in code. https://semver.org/ is ambiguous on the issue. 

I've switched to a minor version upgrade because I think a major version upgrade creates the expectation that users of civisR may need to make changes to their civisR code as a result of this release, which they won't.

patr1ckm:
@mheilman back to you

mheilman:
We could potentially highlight the default change as a breaking change in the changelog to make it a bit clearer that that's the only reason for the version bump. e.g., `- Default docker image for Civis Futures is now `latest` rather than `3`. This is theoretically a change that could require users to update code (#238)`

patr1ckm:
Thanks! Back to you.

PR #242, opened 2020-06-22
auto-doc update

patr1ckm:
This is mostly an autogenerated doc update using the new version of `pkgdown`. There is a new `404.html` page that I added by hand, and new custom js for the toc. All of this was provided by `pkgdown`.

I clicked around on a preview locally and it looked ok.


patr1ckm:
@sahildshah1 in the spirit of keeping the process moving, do you mind taking a look?

sahildshah1:
I clicked around locally and it looked okay as well. I love the new TOC on the right hand side of the screen and the home button in the menu!

Issue #223, opened 2020-01-27
BUG: `fetch_all` throws error on `aliases_list` API function 


sahildshah1:
`template_alias_objects <- fetch_all(aliases_list,object_type='template_script')` 

Produces:

`Error in if (any(sapply(res, .until))) { :  missing value where TRUE/FALSE needed`

Stepping through the [relevant code in pagination_helpers.R](https://github.com/civisanalytics/civis-r/blob/fbea76c36386690846bc9e9d025d235f10f31ede/R/pagination_helpers.R#L76-L78):

```
res <- do.call(aliases_list, list(object_type='template_script'))

str(res)
List of 23
 $ :List of 6
  ..$ id         : int 11
  ..$ objectId   : int 12367
  ..$ objectType : chr "template_script"
  ..$ alias      : chr "civis-ShapefileExport"
  ..$ userId     : int 3001
  ..$ displayName: chr "Export Shapefile"

.until = function(x) x == FALSE

sapply(res, .until)
             [,1]  [,2]  [,3]  [,4]  [,5]  [,6]  [,7]  [,8]  [,9] [,10] [,11]
id          FALSE FALSE FALSE FALSE FALSE FALSE FALSE FALSE FALSE FALSE FALSE
objectId    FALSE FALSE FALSE FALSE FALSE FALSE FALSE FALSE FALSE FALSE FALSE
objectType     NA    NA    NA    NA    NA    NA    NA    NA    NA    NA    NA
alias          NA    NA    NA    NA    NA    NA    NA    NA    NA    NA    NA
userId      FALSE FALSE FALSE FALSE FALSE FALSE FALSE FALSE FALSE FALSE FALSE
displayName    NA    NA    NA    NA    NA    NA    NA    NA    NA    NA    NA


any(sapply(res, .until))
[1] NA

```

sahildshah1:
I think we can fix this bug either by setting the `na.rm` argument to the `any` function to `TRUE`.  
```
any(sapply(res, .until),na.rm=TRUE)
[1] FALSE
```

patr1ckm:
Thanks for the report! This is really helpful.

In this case:

`fetch_until(aliases_list, limit=1000, object_type='template_script', function(x) length(x) == 0))`

Would do the trick without a code change.

For the fix, it's likely that option 1 is fine. Option 2 would be great to have. We'd want to surface `.until` as an argument with the current default as the default value.

sahildshah1:
[...]

PR #243, opened 2022-03-04
Patch fetch_all pagination helper

BryanBaird:
The `fetch_all` function is intended to call `fetch_until` until a result of `FALSE` is returned. However, when the API returns no result, that output manifests as a NULL, not a boolean `FALSE` value. This change prevents a coercion error of trying to boolean compare a NULL value in R.


BryanBaird:
This appears to be related to https://github.com/civisanalytics/civis-r/issues/223, although the approach recommended there is `function(x) length(x) == 0` instead of `function(x) is.null(x)`. I am not sure which other endpoints or result types would need to be tested; perhaps some endpoints could return non-null outputs.

pcooman (approved):
I added you as a contributor. LGTM!

PR #248, opened 2023-02-10
Replace deprecated ggplot2 functions in CivisML plots

pcooman:
* replaces `ggplot2::aes_string()` (deprecated) with `ggplot2::aes()` and uses the symbolic column names instead of string names
* replaces the deprecated `size` argument in `ggplot2::geom_hline()` with `linewidth`

### DIFF
--- a/R/civis_ml_plot.R
+++ b/R/civis_ml_plot.R
@@ -27,7 +27,7 @@ hist.civis_ml <- function(x, name = NULL, ...) {
   if (!is.null(name)) df <- subset(df, names == name)
 
   ggplot2::ggplot(df) +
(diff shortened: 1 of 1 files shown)


mattyb:
It looks like `linewidth` was only introduced in 3.4.0 https://ggplot2.tidyverse.org/news/index.html#ggplot2-340. Do we need to enforce that as a dependency?

mattyb (changes requested):
See above comment re: pinning the dependency

pcooman:
Let's hold off on this for now. ggplot2 v3.4.0 was released only recently (11/2022), so a lot of people won't have upgraded to yet. Also, Platform itself is still using ggplot v3.3.3.

Issue #245, opened 2023-01-18
future: Rename 'state' from `"succeeded" to "finished"


HenrikBengtsson:
Hi, I'm doing more cleanups on the **future** package and I noticed that you use `"succeeded"` to denote that a CivisFuture has finished successfully, e.g.

https://github.com/search?q=repo%3Acivisanalytics%2Fcivis-r%20succeeded&type=code

However, elsewhere in the future ecosystem, we use `"finished"` for this state.  Could you please update your package accordingly?   Looking at the above search, it could be that:

```sh
$ sed -i 's/"succeeded"/"finished"/g' R/*.R tests/testthat/*.R
```

is all that you need to do.

Updating this would help me remove [a hard-coded special case for **civis**](https://github.com/HenrikBengtsson/future/blob/d1218528b26871ba8d5bbf5d666d266adcf781e9/R/signalConditions.R#L29-L30) in the **future** package;

```r
  ## FIXME: civis::CivisFuture uses 'succeeded' /HB 2019-06-18
  if (!future$state %in% c("finished", "failed", "succeeded")) {
    ...
  }
```

Thank you

pcooman:
Hi @HenrikBengtsson,

Unfortunately, the "succeeded" state does not come from `CivisFuture()` itself, but from the Civis API. The possible states for a job in the Civis API are: "idle", "queued", "running", "succeeded", "failed", "cancelled", and "sleeping". To fully replace "succeeded" with "finished", we would not only need to replace it here in the `civis-r` repo (for the R Client), but also in the Civis API, and in the other Clients (for Python and JS). 

Also, this change is highly likely to break user pipelines: a child job may be waiting for a "succeeded" response from a parent job before executing.

We would greatly appreciate if you could keep the exception in place! Would that be possible? If not, there are a few things we could do:
1. Upper bound the version of the `future` package in the `civis` DESCRIPTION file. We currently Import `future (>= 1.8.0)`, so we could add something like `future (<= 1.31.0)` to make sure our users only use the `civis` package with a `future` package that still has the Civis exception.
2. We could try to add a copy of the `future::SignalConditions()` function with the Civis exception into `civis` and redirect our `future::Future()` to use the `civis::SignalConditions()` function instead?

Do you have a preference? Or maybe another suggestion for how to resolve this issue?

HenrikBengtsson:
Update

```
    future$state <- scripts_get_containers_runs(id = future$job$containerId,
```

so that the correct string in stored.

[...]

PR #251, opened 2023-02-21
Set future$state to "finished" if API call "succeeded"

pcooman:
New versions of `future` (> 1.31.0) will no longer recognize "succeeded" as a resolved state. This PR sets `future$state` to "finished" if the API job has a "succeeded" status.

See also Issue #245


mheilman (approved):
LGTM

PR #255, opened 2025-01-16
Fix CI build and update API docs

mheilman:
This fixes the CI build. Details:

* Updates `.travis.yml` to use Docker with the same public image from rocker that we build [datascience-r](https://github.com/civisanalytics/datascience-r) from. This also adds little script to install the package and run checks and tests (tools/install-and-run-tests.R).
* Updates the Roxygen version in the DESCRIPTION file since there's a newer version in rocker 4.4.
* This rebuilds the docs from the API spec with a fix for an issue in the Rd documentation files related to brace characters that was flagged when running `devtools::check()`. I think the new version of Roxygen flagged the issue.
* Since the docs were rebuilt, this adds some new endpoints (e.g., for IDR and dbt), and removes some admin ones that shouldn't have been included.


mattyb (approved):
LGTM

Issue #254, opened 2025-01-10
`with_mock()` is going away


hadley:
This is a heads up to let you know that `with_mock()` is deprecated and will be made defunct soon. The API functions that made it work are being removed in R 4.5.0, so we'd suggest that you switch to `with_mocked_bindings()` instead.

mheilman:
I'm closing this since #256 removed `with_mock()` and replaced it with `local_mocked_bindings()`, which seemed a bit easier to use than `with_mocked_bindings()`. Please feel free to reopen it if you still have concerns.

Thanks for filing this!

PR #256, opened 2025-01-16
Replace `with_mock` with `local_mocked_bindings` in the tests

mheilman:
This replaces [`with_mock()`](https://testthat.r-lib.org/reference/with_mock.html) with [`local_mocked_bindings()`](https://testthat.r-lib.org/reference/local_mocked_bindings.html) since the former is deprecated and won't be available in R 4.5. See #254.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ Meta/
 CRAN-SUBMISSION
 cran-comments.md
 .Renviron
+..Rcheck
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,11 @@
 os: linux
 dist: jammy
 language: generic
-# Test this recent and not so recent versions of R.
+# Test this with multiple versions of R.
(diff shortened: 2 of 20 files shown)


lhoesly:
Adding a small note that I did get 4 warnings when I ran all the tests locally but they seem to be unrelated to these changes

mheilman:
@lhoesly, thanks for mentioning the warnings. I also see warnings locally, for this and the master branch, so I think it's an issue for some future PR.  I see 5 rather than 4, perhaps because the version or R or some packages are different than on your local machine.

In [90be49a](https://github.com/civisanalytics/civis-r/pull/256/commits/90be49ad346af110ddc0f71b91a21c8f71884454), I updated the checks in CI so that the tests are run separately from other checks, and so the warnings get recorded. Could you please take another look when you get a chance?

lhoesly (approved):
LGTM!

PR #259, opened 2025-08-25
Prerelease 3.1.3

dleatherman-civis:
Updates that need to be made as a part of the release process.

Once this is merged, I will submit to CRAN

PR #260, opened 2025-08-28
Prerelease Fix 3.1.3

dleatherman-civis:
A follow up to #259. This restores the escape characters. It isn't clear why they were omitted in the first place but I suspect it is due to local environment vagaries. There are some documentation NOTES from `devtools::check()` that I addressed manually because I wasn't sure how to do it in an automated fashion.

This package has been built and successfully published to CRAN.


lhoesly (approved):
LGTM!

